PR #9, opened 2019-08-19
Identification of instances of a type

SimonShapiro:


elf-pavlik:
It sounds to me more like a specific requirement, some use case could illustrate scenario which justifies that specific requirement. Use case preferably would include some actors which have some agency, for example specifying preferred location for resources (documents) with information about things of specific type. One use case in many cases will lead to multiple requirements.

RubenVerborgh:
Indeed; creating something of a particular type is a solution to a problem; we need that problem.

SimonShapiro:
Ah.  Usually data of a particular type has a predictable shape - often described by shacl.  UI's could be generated from the shacl to allow a user to capture an instance of a particular type.  For an app to store this on a user's pod, the location, index mechanism and identification mechanisim is needed.  Otherwise, the user is left to make all these decisions.

Is each instance a document, is it a fragment in some other document?  How is the IRI of the document/fragment to be created.

RubenVerborgh:
>  Usually data of a particular type has a predictable shape

No no, these are completely distinct. Hence why we have shapes in the first place.

SimonShapiro:
ok. So is a shape a template of triples in which an app has a particular interest?

RubenVerborgh:
For instance, yes!

SimonShapiro:
That still leaves the problem of identification, or IRI generation.  Say, I want to store copies of my bank transactions on my POD, there is no guarantee the bank will provide nice clean IRIs for each transaction.  Usually they will give you a csv or some other interchange format.  So I need to know how to generate IRIs.  If I allow users to send a photo to me via email, again, how do I arrange matters for the POD to be able to identify them?

If you say, well the POD can generate anything and then all incoming information is just predicate-objects in the triples, the problem moves to where the user is switching from one app to another.  How do we perpetuate the id scheme in each app?  

Or are we giving up on that and saying that the POD controls location (via the `slug`), tells the app what it has decided and the app then has to react to that by placing the generated IRI in somesort of index.  And at what grain does that happen?

SimonShapiro:
> > Usually data of a particular type has a predictable shape
> 
> No no, these are completely distinct. Hence why we have shapes in the first place.

From [the shacl spec](https://www.w3.org/TR/shacl/#targetClass)

> If s is a shape in a shapes graph SG and s has value c for sh:targetClass in SG then the set of SHACL instances of c in a data graph DG is a target from DG for s in SG.

RubenVerborgh:
Yes, shapes can apply to certain classes (types). But that does not mean that a type has a predictable shape.

RubenVerborgh:
> That still leaves the problem of identification, or IRI generation.

Yes, it does. So we should phrase those as high-level problems. Like types, IRI generation is a solution to a problem, and that problem should be documented as use cases.

> Or are we giving up on that

I definitely wouldn't; shapes are the way to go.

SimonShapiro:
I added an example shacl file based on Markbook's file structure.  I use it as an example of the challenges faced by app providers:  How closely should my app follow this shape? How do I identify groups of triples that validate against this shape - that is, what is the IRI of the subject?  Do I store many in one document, or each in its own document?

SimonShapiro:
So a use case could be:

> As an app provider, how and where should I record the shape that my app uses to generate the UI and validate data discovered on a users's POD.

justinwb:
> So a use case could be:
> > As an app provider, how and where should I record the shape that my app uses to generate the UI and validate data discovered on a users's POD.

Closer, but we're conflating a few things together here (what shapes you use, where to write the shape, generating ui from a shape, and validation vs. discovered data).

As far as format, there are a number of formats we can follow for use cases, but here's a [good example](https://www.w3.org/TR/verifiable-claims-use-cases/) from the Verifiable Credentials W3C track.

SimonShapiro:
My experiment continues...

I will write-up my apporach in more detail in the next round.  But roughly,

My app brings in it's own shape ttl.  Based on this it selects all triples matching the shape and displays some of them in the ui; hiding the creator and date created, for example.  I now need to decide on how to handle a delete command.  Am I deleting:

- the displayed triples
- all the triples in the shape
- all the triples for that particular resource?

My instinct is for the second option, any other suggestions or comments?

RubenVerborgh:
Interesting and relevant question. I'm about to make it a bit more difficult 🙂

> My app brings in it's own shape ttl. Based on this it selects all triples matching the shape and displays some of them in the ui;

Strictly speaking, if an app is not going to do anything with part of a shape, it will likely not include those in the shape. So if creator and date created are not part of the app, I don't see why they would be in the shape. (Unless they would be used elsewhere of course, such as for writing.)

> all the triples for that particular resource?

The problem is that this selection is ill-defined. There is no clear way to determine what those triple are.

SimonShapiro:
> all the triples for that particular resource

I guess, more precisely:

```
delete where {
<my_subject> ?p ?o
}
```
I still think that the second option is better: `all triples in the shape`.  In the example quoted, `DateCreated` is not displayed on the ui, but used for sorting a list of bookmarks.

As an aside, should the shape always include a `type` property?

```
:BookmarkShape
    rdf:type rdfs:Class;
    rdf:type sh:NodeShape;
    sh:targetClass bookm:Bookmark;
    sh:property [
        sh:path rdf:type;
        sh:description "The type property for the bookmark";
        sh:name "Bookmark Type";
        sh:hasValue bookm:Bookmark;
    ];
```

RubenVerborgh:
That query would leave dangling blank nodes and other subordinate
resources. My point being that I don’t think option 3 can be implemented,
so we can remove it.

I’m also interested in option 2 indeed.

Shapes don’t always need a type IMHO. (Certainly for reading, it can be
inferred.)

data-discovery/use-cases.md
```diff
@@ -2,12 +2,19 @@
 
 The following use cases must be considered when designing a mechanism for a data pod to define how its data is stored and indexed so that a given application can automatically find the data that it needs.
 
-## Identification scheme
+## New application
 
-In order for data of a particular type to be created,
-we need to tell an application what the prefered mechanism for creating
-instance ids should be.
+Alice used `AncientBookmarks` application from `OldBookmarks.com`, to record a number of bookmarks which are stored on her POD.  Recently Alice's friend Bob recommended that she upgrade to the application `MyThoughtsOnIt` from `WebAnnontaions.com`.
```
TallTed: Conformance to [RFC2606](https://tools.ietf.org/html/rfc2606#section-2) suggests that `OldBookmarks.com` and `WebAnnontaions.com` be changed to `OldBookmarks.example` and `WebAnnontaions.example`, respectively, here and subsequently in this document.

data-discovery/use-cases.md
```diff
@@ -15,9 +15,9 @@ whereas `AncientBookmarks` used a proprietary ontology to define its data.
 
 Alice needs to authorise `MyThoughtsOnIt` to upgrade and possibly move the bookmarks created by `AncientBookmarks`.
 
-#### WebAnnotations.com
+#### WebAnnotations.exmaple
```
TallTed: typo -- `exmaple` -> `example`

SimonShapiro:
Fixed typo and moved approach to hypothesis.md in order to allow the use case to go ahead independently of the hypotheses.

elf-pavlik:

PR #11, opened 2019-08-20
Added read-time shape validation use case.

jaxoncreed:


jaxoncreed:
> Find data associated with a given shape in a resource that has a lot of other data in it

I'd say that's already prescribing a solution isn't it? Shouldn't use cases be implementation agnostic. Maybe this use case does have some authorization in it, but the data interop use cases should be able to have overlap with other subjects.

data-discovery/use-cases.md
```diff
@@ -1,3 +1,5 @@
 # Use Cases for Data Discovery
 
 The following use cases must be considered when designing a mechanism for a data pod to define how its data is stored and indexed so that a given application can automatically find the data that it needs.
+
+ - Alice has a super file with a wide array of data that falls under many healthcare-related shapes on her Pod. She wants to share only data about ultrasound readings with her new OB-GYN without sharing information on her skin care, dentistry or other subjects. (Read-time shape validation)
```
elf-pavlik: I don't see reason why use case should assume 'a super file'. Preferably it will just state 'a wide array of data' and leave it open to solution where data needs to stay in separate resources to manage access to them separately.

elf-pavlik:

elf-pavlik:
I messed up this branch but the initial PR should be in https://github.com/solid/data-interoperability-panel/commit/e1dcd3cc323deab8abd5c05541a7583153ea9d80

PR #14, opened 2019-08-21
A few to start with.

JKReynolds:
**These are some proposed use cases**
Job Applications
Resume Builders
Calendar Applications
Message Applications
Car insurance Applications
Apartment Rental Applications
Loan Applications


JKReynolds:
My team and I put together a couple use cases that we would like to get some feedback on. Please let us know what you think. Thanks.

dmitrizagidulin:
While we're talking about this, I do want to bring up PR https://github.com/solid/identity-panel/issues/1 which (among other things) proposes that we should separate 'user info' type data (most of which is private / access-controlled) from the public WebID profile doc.

dmitrizagidulin:
(Which also echoes what @justinwb is saying in the previous comment, "information that should be discoverable via someone's user profile, not contained inside of it")

mark-jensen:
@justinwb @dmitrizagidulin  I am not sure I understand the distinction between data being discoverable via a user profile vs. being contained in it. The distinction seems to track one between public vs private, but if the private data is merely discoverable via a user profile, where is it "contained"?

mark-jensen:
We have posted initial components for the User Profile use case described in this PR ([link](https://github.com/solid/data-interoperability-panel/blob/49bf3b2debfbe20af6fe97b9fc0dee4b9558300c/user-profile/use-cases.md)), as well as in #24, to a project repository located [here](https://github.com/CommonCoreOntology/UserProfileOntology)

Thus far we have:

- User Profile Ontology module which extends CCO
- Sample batch of RDF for a profile
- An example of a shape for person's full name and nickname
- A spreadsheet of relevant ontology terms and definitions

Some members of our team will attend Monday's (10/21) Data Interoperability Panel call to provide a brief overview, discuss and answer questions.

elf-pavlik:

PR #22, opened 2019-08-28
change dates to ISO-standard YYYYMMDD

TallTed:


TallTed:
@justinwb - Done. Hadn't realized that external linkage was there!

PR #23, opened 2019-09-09
Initial round of use cases for data validation

justinwb:
Use cases for common problem scenarios, then adding, processing, and modifying validations.

### DIFF
--- a/data-validation/use-cases.md
+++ b/data-validation/use-cases.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-Server-side data validation is an essential factor to the seamless interoperability promised by Solid. Multiple applications that manipulate the same kind of data, developed with no knowledge of each other, should be able to write that data without breaking each other. This requires all applications to conform to a common schema. Data Shape languages like [ShEx](http://shex.io/shex-semantics) and [SHACL](https://www.w3.org/TR/shacl/) help us solve this.
+Server-side data validation is an essential factor of the seamless interoperability promised by Solid. Multiple applications that manipulate the same kind of data, developed with no knowledge of each other, should be able to write that data without breaking each other. This requires all applications to conform to a common schema. Data Shape languages like [ShEx](http://shex.io/shex-semantics) and [SHACL](https://www.w3.org/TR/shacl/) help us solve this.
 
 Shapes allow us to define schema models with attributes from different linked data vocabularies. We can then test input data against those shape schemas and determine whether that input conforms to the shape or not. In other words, if we have a shape for a comment, and an application tries to write a comment, we can use that shape to validate whether that comment adheres to our expectation (the shape) of how a comment should be structured. Now we can use many different applications to safely read and write comments, just by adhering to the comment shape schema.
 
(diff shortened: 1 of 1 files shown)


data-validation/use-cases.md
```diff
@@ -2,15 +2,148 @@
 
 ## Introduction
 
-Server-side data validation is an essential factor to the seamless interoperability promised by Solid. Two applications that manipulate the same kind of data, developed with no knowledge of each other, should be able to write that data without breaking each other. Using common linked data vocabularies is not enough to satisfy this fundamental need. Data Shape languages like [ShEx](http://shex.io/shex-semantics) and [SHACL](https://www.w3.org/TR/shacl/) help us solve this.
+Server-side data validation is an essential factor to the seamless interoperability promised by Solid. Multiple applications that manipulate the same kind of data, developed with no knowledge of each other, should be able to write that data without breaking each other. This requires all applications to conform to a common schema. Data Shape languages like [ShEx](http://shex.io/shex-semantics) and [SHACL](https://www.w3.org/TR/shacl/) help us solve this.
```
TallTed: `factor to the seamless`
-> `factor of the seamless`

data-validation/use-cases.md
```diff
@@ -2,15 +2,148 @@
...
+Shapes allow us to define schema models with attributes from different linked data vocabularies. We can then test input data against those shape schemas and determine whether that input conforms to the shape or not. In other words, if we have a shape for a comment, and an application tries to write a comment, we can use that shape to validate whether that comment adheres to our expectation (the shape) of how a comment should be structured. Now we can use many different applications to safely read and write comments, just by adhering to the comment shape schema.
 
-Unfortunately, we cannot trust that all clients will write conformant data on their own. Consequently, we must be able to enforce conformance to a given shape on-pod, as the client writes the data.
+Schema validators are an enormous aid to application developers, but we cannot trust that all clients will write conformant data all the time. Consequently, we must be able to enforce conformance to a given shape at the pod, as the client writes the data.
+
+## Problem Cases
+
+The following use cases detail some of the problems related to a Solid ecosystem without on-pod data validation.
+
```
TallTed: The bullets in each problem case appear to be a sequence of events, the order of which is important. They should therefore be numbered (ordered) lists, instead of bulleted (unordered) lists.

data-validation/use-cases.md
```diff
@@ -2,15 +2,148 @@
...
+- The data is migrated, and Alice immediately realizes that two other important applications, TrueHealth and YouHealth, no longer work.
+
+__An application prompts a user to make decisions about shape constraints__
+- Bob has a brand new data pod, and has recently decided to use ExampleApp to manage the data he's collecting as part of research and development work that he's performing.
+- ExampleApp works with a wide variety of new shapes that Bob hasn't stored before, and writes data to the pod that conforms to different combinations of these shapes depending on the workflows Bob uses.
+- Before ExampleApp starts writing data, it prompts Bob and asks if he would like to enforce shape constraints on this data in his pod.
+- Based on the shapes it will write data for, and the places it will store them, ExampleApp prompts Bob with a bunch of toggles to determine which ones he does and doesn't want to enforce.
+- Bob has no idea what shapes are. He get intimidated and abandons ExampleApp and Solid.
 
 ## Use Cases
```
TallTed: As above, the bullets in each use case appear to be a sequence of events, the order of which is important. They should therefore be numbered (ordered) lists, instead of bulleted (unordered) lists.

data-validation/use-cases.md
```diff
@@ -2,15 +2,148 @@
...
+- Alice's doctor uses an EHR system called BarEHR that is Solid compatible.
+- BarEHR maintains a care plan resource at `<https://alice.example/phr/care-plan.ttl>` that maps to a number of other resources such as prescribed diagnostics, prescriptions for medication, and other required activities.
+- This care plan resource has a validator set that ensures it conforms to `<https://shapes.example/care-plan.shex>`.
+- A new version of BarEHR allows it to canonicalize and sign the care plan data, then store the results in the same resource, conforming to the `<https://shapes.example/data-signature.shex>` shape.
+- BarEHR adds `<https://shapes.example/data-signature.shex>` to the allowed validators for `<https://alice.example/phr/care-plan.ttl>`, such that both must be present (the care plan is always signed) to pass validation.
+
+#### Adding additional shape validations to a container
+
+- Bob uses FooChat to communicate with his co-workers.
+- Bob creates a chatroom where he can talk to his colleague Rob.
```
elf-pavlik: Suggestion not specific to this PR, same use case will apply to other projects than validation, we may need some base use case and then project specific aspects of the same use case. Direct chat can have variants where chatroom gets created in organization storage or one or both of personal storage. It can also have requirements like data redundancy eg. when it gets created in Bob's personal storage, Rob wants to have copy of all messages in his storage for redundancy - Bob's storage going offline shouldn't cause Rob to loose chat history. I see it useful to have all chat related use cases organized together and then explore different aspects of them in various projects.

data-validation/use-cases.md
RubenVerborgh: I actually do not agree with the phrasing above.

If all applications conform to a common schema, then we need centralized agreement, and can solve it with much easier tooling than RDF.

The key challenge is how to interoperate in absence of slow central agreement processes.

data-validation/use-cases.md
```diff
@@ -2,15 +2,148 @@
...
+- Bob creates a chatroom where he can exchange messages with Alice and Mary.
+- FooChat writes data conforming to a chat shape at `<https://bob.example/chats/maryalice.ttl>`, and turns on shape validation for `<https://bob.example/chats/maryalice.ttl>` to ensure that data in this resource conforms to the Chat shape.
+
+#### Enabling shape validation on a container
+
+- Bob uses FooChat to communicate with his co-workers.
+- Bob has a chatroom where he can exchange messages with Alice and Mary.
+- When Bob created the chatroom, FooChat setup a container to store messages.
+- The container organizes messages into sub-containers by month and day
+  - `<https://bob.example/chats/maryalice/2019/8/1.ttl>`
```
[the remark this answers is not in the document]
justinwb: We discussed this live in the panel meeting but recording here before resolving. Data discovery and footprints should be used to find things / identify where to write things and that's outside of the scope of validation so not trying to solve it here (though it will be solved elsewhere)

data-validation/use-cases.md
justinwb: Agree that we should have top level use cases for Solid. Will propose as a broader issue outside of this PR.

Issue #10, opened 2019-08-19
Discoverability of root and controllers of Pods - some thoughts


megoth:
This is probably a bit too early in the process, but wanted to share some thoughts for a couple of data points that we've had to assume in the current data browser (as the current spec doesn't help us).

I've written up the thoughts on https://megoth.inrupt.net/public/SolidDataBrowser/pod-info.html. If you use dokieli to annotate, please let me know, so I can add you to my WebID as `foaf:knows` (which should enable me to see your annotations).

I would hope that we can standardize some mechanism that allows for discoverability of storage for a given resource. I also think discoverability of controllers are useful, but I understand that there are tougher questions to deal with on that one.

RubenVerborgh:
(Sorry, pressed the wrong button.)

RubenVerborgh:
I don't think "root of pod" should be a thing; I don't think the concept of "pod" has a place in the data model. It's only resources and folders.

megoth:
> I don't think "root of pod" should be a thing; I don't think the concept of "pod" has a place in the data model. It's only resources and folders.

If all of these instances are replaced with `storage`, does that make it better?

The goal is to have some way of knowing what the storage for a given resource is without assuming anything about the setup for the given Pod.

RubenVerborgh:
> If all of these instances are replaced with `storage`, does that make it better?

Probably, yes.

> The goal is to have some way of knowing what the storage for a given resource is

Or in the solution-oriented (which is not always good) terminology I currently have in mind: what shape it conforms to. So then the resource would be part of a folder, which would conform to a shape structure, recursively.

megoth:
> > The goal is to have some way of knowing what the storage for a given resource is
> 
> Or in the solution-oriented (which is not always good) terminology I currently have in mind: what shape it conforms to. So then the resource would be part of a folder, which would conform to a shape structure, recursively.

I agree that shapes are important for resources, but that doesn't solve the question of which storage a resource belongs to, I think? Just asking in case I have missed something wrt shapes.

RubenVerborgh:
> doesn't solve the question of which storage a resource belongs to, I think?

I don't think that question really has an answer though. A resource might be associated with zero, one, or multiple storages.

megoth:
> > doesn't solve the question of which storage a resource belongs to, I think?
> 
> I don't think that question really has an answer though. A resource might be associated with zero, one, or multiple storages.

[...]

PR #32, opened 2019-12-18
In-progress draft of Resource metadata (FOR COLLABORATIVE EDITING)

justinwb:
Per the discussion in the [most recent meeting](https://github.com/solid/data-interoperability-panel/blob/master/meetings/8-20191216.md#minutes) of the data interoperability panel, I'm submitting this **incomplete and very much still in-process** draft of resource meta-data so that panel members can provide some feedback and deliberate on the content.

Please do not mistake this pull request as a representation that this content in its current form is considered ready to submit to the specification. The purpose of the pull is to give us an easy place on github to collaborate and help get it there.


kjetilk:
I've realized that I don't understand how it currently works, and I still miss some information to understand it.

The main thing I fail to understand relates to the question: "what is a resource in its own right?"

I've observed that if you add triples to a container, those triples will be added to `.meta`. In that case, `.meta` is not a resource in its own right, I suppose. You wouldn't interact with it, you'd interact with the container. 

Then, I suppose that if you have an image, and you wish to add metadata to the image, you would do that to a separate resource, and you would get metadata by dereferencing a separate resource. So, in that case, the metadata resource is a resource in its own right. Concievably, you could discriminate between the representations by media type though.

It would be nice to have clarification around that.

kjetilk:
BTW, I believe we should get to a position where we can check of @dmitrizagidulin 's boxes in https://github.com/solid/specification/issues/63

resource-metadata/README.md
```diff
@@ -0,0 +1,7 @@
+# Resource Metadata in Solid
+
+A project undertaken by the [Data Interoperability Panel](https://github.com/solid/data-interoperability-panel).
+
+## Goal
+
+The aim of this project is to produce a candidate proposal to the [Solid Specification](https://github.com/solid/specification) that defines a mechanism for how metadata (i.e. access control, shape constraints) is associated and accessed for a given resource.
```
TallTed:
```suggestion
The aim of this project is to produce a candidate proposal to the [Solid Specification](https://github.com/solid/specification) that defines a mechanism for how metadata (such as access control or shape constraints) is associated and accessed for a given resource.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,121 @@
...
+
+Link: <https://example.com/resource?meta>;
+        rel="https://example.org/ns#hasMetadata"
+
+and:
+
+Link: <https://example.com/resource>;
+        rel="https://example.org/ns#isMetadataOf"
+
+String-based link relations, such as in the examples above, must be registered with IANA. But it is also possible to use custom relation types by using a full IRI. The Linked Data Platform, the Linked Data Notification and the Web Annotation specifications make use of full IRIs in the rel attribute.
```
TallTed:
```suggestion
String-based link relations, such as in the examples above, must be registered with IANA. But it is also possible to use custom relation types by using a full IRI. The [Linked Data Platform](https://www.w3.org/TR/ldp/), the [Linked Data Notifications](https://www.w3.org/TR/ldn/), and the [Web Annotation Protocol](http://www.w3.org/TR/annotation-protocol/) specifications make use of full IRIs in the `rel` attribute.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,121 @@
+# Resource Metadata Model
+
+## Background
+
+This document introduces a mechanism for linking to metadata about resources in the Solid Ecosystem using HTTP Link headers. Examples of this linking mechanism include:
+
+- A binary JPEG image with a Link header, pointing to a description that has an RDF representation.
```
elf-pavlik: I have created dedicated issue for media resources and multiple encodings of same creative work #33 . Here I would like to clarify if we talk about description of each of those encodings or description of the creative work which can have numerous encodings?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Implementation Patterns
+
+There are many ways a server could implement these features. A file-based server could have a special naming scheme reserved for these metadata resources. Alternatively, a server could represent every resource internally as a dataset, storing each separate type of metadata in its own named graph.
+
+[ Include example of file based ]
+
+[ Include example of graph based ]
+
+## Open Questions
```
elf-pavlik: I think we should create dedicated issue of each of those questions.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,121 @@
...
+
+Authorization Model:
+
+Discovered via ```rel=http://www.w3.org/ns/solid/terms#shape``` and ```rel=http://www.w3.org/ns/solid/terms#resource```
+
+### Server Managed
+
+Server-managed metadata not otherwise included in HTTP headers. MUST be supported.
+
+__Note: This may be too broad to be able to properly determine what the authorization rule, or lifecycle should be. May need to consider breaking this up into specific types.__
```
elf-pavlik: Since we don't expose them over HTTP I don't think authorization rules apply to them. We may just need to define something to make them portable.

justinwb: > I've realized that I don't understand how it currently works, and I still miss some information to understand it.
> 
> The main thing I fail to understand relates to the question: "what is a resource in its own right?"
> 
> I've observed that if you add triples to a container, those triples will be added to `.meta`. In that case, `.meta` is not a resource in its own right, I suppose. You wouldn't interact with it, you'd interact with the container.
> 
> Then, I suppose that if you have an image, and you wish to add metadata to the image, you would do that to a separate resource, and you would get metadata by dereferencing a separate resource. So, in that case, the metadata resource is a resource in its own right. Concievably, you could discriminate between the representations by media type though.
> 
> It would be nice to have clarification around that.

Take a look at latest changes, which stipulate that metadata resources on the same server follow the LDPR interaction model, and then (in implementation patterns) that servers know/maintain which resources are metadata, and can apply special handling to them.

justinwb:
> I've realized that I don't understand how it currently works, and I still miss some information to understand it.
> 
> The main thing I fail to understand relates to the question: "what is a resource in its own right?"

@kjetilk Take a look at latest changes, which stipulate that metadata resources on the same server follow the LDPR interaction model, and then (in implementation patterns) that servers know/maintain which resources are metadata, and can apply special handling to them.

justinwb:
> BTW, I believe we should get to a position where we can check of @dmitrizagidulin 's boxes in [solid/specification#63](https://github.com/solid/specification/issues/63) too

Agree - I've added a temporary section at the bottom where I went through every related issue I could find, and then extracted items / points that need to be addressed. #63 was contributed to most of them.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,180 @@
...
+
+A Solid server SHOULD sanity check Shape validation metadata resources upon creation or update to restrict invalid changes.
+
+### Server Managed
+
+*This section is still in process*
+
+Server-managed metadata not otherwise included in HTTP headers. MUST be supported.
+
+Discovered via ```rel=http://www.w3.org/ns/solid/terms#server``` and ```rel=http://www.w3.org/ns/solid/terms#resource```
```
elf-pavlik: If server doesn't advertise them in HTTP Link Header where would those link relations appear and who would try to perform discovery based on them?

resource-metadata/hypothesis.md
elf-pavlik: If we just talk about different representation of the same resource, I think we could just use `rel=alternate` together with `type=image/jpeg` http://microformats.org/wiki/rel-alternate#With_type

> When used in combination with the `type` attribute (with a value other than that of the document itself; e.g. other than `"text/html"`), `rel="alternate"` means a link to a representation of the contents of the current document in a different format, as designated by the `type` attribute. 

If we talk about relating a creative work to multiple different encodings, each of which can have their own encoding specific rdf representation, i think should discuss it in #33

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Shape Validation
+
+Shape Validation ensures that any data changes in a given resource conform to an associated [SHACL](https://www.w3.org/TR/shacl/) or [ShEx](https://shex.io/shex-semantics/index.html) data shape. It MUST be supported as a resource metadata type by Solid servers.
+
+The Shape validation metadata resource for a given resource is discovered via ```rel=http://www.w3.org/ns/solid/terms#shape``` and ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+.```OPEN QUESTION - Should the link be directly to the shape resource (which is often not going to be on the same server)? Should it be to a local resource that in turn points to the shape (wherever it is), but can also include configuration parameters on the nature of the validation? The latter would seem to provide much more flexibility in the long run.```
```
elf-pavlik: As I understand, server has full and exclusive control when it comes to creating all the 'metadata resources' and links between them and the resource. Client would not have any direct control over those link relations and could only have to those metadata resources. In that case for client to have possibility to reference some specific shape that resource should validate against, it would need to modify specific metadata resource linked by the server with `solid:shape` relation.

elf-pavlik:
https://github.com/solid/specification/issues/58#issuecomment-534634023
> 2) If a resource is deleted and its .acl remains, the .acl is undiscoverable and its container undeleteable 

Besides discussions like in https://github.com/solid/specification/issues/131 it seems that server have full control over assigning location of acl resource for any protected resource. In that case server can choose very specific location for it eg `/.well-known/solid-meta/wac/{uuid}` or `/.well-known/solid-meta/wac{resource_path}` (anything that clients would not try to use to create non meta data resources). This way acl resources could get treated as regular LDP Resources including containment etc. Personally I find adding notion of 'hidden / unlisted' resources carrying some design smell. For example if server just appends `.acl` it has to prevent clients from creating any non acl resource with path ending in `.acl`, this would especially affect handing creation with PUT and may require special error messages like 'path reserved for server created metadata'.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,180 @@
...
+Some examples include:
+
+- An ACL resource controls how a server makes authorization decisions for a container and any child resources.
+- A shape constraint resource may limit the RDF structures that can be added to a resource or container.
+- A “configuration” resource may control how a resource is versioned, or which indexes are exposed for it.
+
+Access to different categories of metadata may require different levels of privilege, which must be specified as part of the definition for a given metadata type.
+
+## Metadata Discovery
+Given the URL of a resource, a client can discover the metadata resources by issuing a GET (or HEAD) request and inspecting the Link headers in the response. The [rel= attribute](https://tools.ietf.org/html/rfc8288) will define the relationship to the target URL.
```
TallTed:
```suggestion
Given the URL of a resource, a client can discover the metadata resources by issuing a GET (or HEAD) request and inspecting the Link headers in the response. The [`rel={attribute}`](https://tools.ietf.org/html/rfc8288) will define the relationship to the target URL.
```

kjetilk:
> > I've realized that I don't understand how it currently works, and I still miss some information to understand it.
> > The main thing I fail to understand relates to the question: "what is a resource in its own right?"
> 
> @kjetilk Take a look at latest changes, which stipulate that metadata resources on the same server follow the LDPR interaction model, and then (in implementation patterns) that servers know/maintain which resources are metadata, and can apply special handling to them.

It doesn't quite address my problem, the behaviour I see, in the case of directories, for example, is that `/.meta` contains the triples of the container. There is no discovery involved, I would guess, since the `/.meta` is manipulated through the container, not its own URL. 

I'm also in similar kind of doubt about other media types, you could in principle manipulate the XMP or IPTC metadata in images through manipulations on the RDF. Currently, it is rather counter-intuitive to me that it should be done that way, but these are some expectations that should be researched.

Also, note that LDP interaction models doesn't have an important place in Solid. In particular, there is no requirement on the client to supply a `Link` header to identify the expected interaction model, so the metadata mechanism shouldn't be too tightly coupled to LDP interaction models.

elf-pavlik:
> It doesn't quite address my problem, the behavior I see, in the case of directories, for example, is that `/.meta` contains the triples of the container. There is no discovery involved, I would guess, since the `/.meta` is manipulated through the container, not its own URL.

In that case use of `/.meta` sounds like implementation internal detail. On the other hand, for sake of portability we might want to specify that exported container should include all the statements, no matter how given implementation chooses to persist them in filesystem or elsewhere.

kjetilk:
> In that case use of `/.meta` sounds like implementation internal detail. On the other hand, for sake of portability we might want to specify that exported container should include all the statements, no matter how given implementation chooses to persist them in filesystem or elsewhere.

Indeed, in that case it is an internal issue, so that's what I need to understand, when is it an internal implementation detail, and when is it a resource in its own right.

justinwb:
> > In that case use of `/.meta` sounds like implementation internal detail. On the other hand, for sake of portability we might want to specify that exported container should include all the statements, no matter how given implementation chooses to persist them in filesystem or elsewhere.
> 
> Indeed, in that case it is an internal issue, so that's what I need to understand, when is it an internal implementation detail, and when is it a resource in its own right.

Any reference to ".meta" is an implementation detail to me. Specifically, .meta was a place to stuff metadata for a given resource (in this case a container) on the filesystem in node-solid-server. The goal here is to look a level above that where a given resource (binary, rdf, or container) is able to identify where to look for other types of metadata associated with it. Consequently, I believe that we need to treat all of these as resources in their own right, and leave it up to the servers to know when something is meta (if its on the same server), and add special handling accordingly.

resource-metadata/hypothesis.md
justinwb: Would you say that the proposed approach we describe here gives us the basic capabilities needed to make these associations (for example in the content of the resource description metadata resource)?

resource-metadata/hypothesis.md
justinwb: Per the discussion in the panel meeting yesterday this was poorly worded and a bit misleading as written because it was incomplete. The next round will complete the section and address this so will resolve for now.

justinwb:
Another [update pushed](https://github.com/solid/data-interoperability-panel/pull/32/commits/17790041f9fa6237afadc8b717d4de3ad9391124) - I think this is close to a full draft now. Included descriptions for server managed and configuration metadata types. Clarified interaction model points, and cleared open questions. Note that two issues were created from the open questions: #36, #37

resource-metadata/hypothesis.md
justinwb: Resolve in most recent version.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Reserved Metadata Types
+
+### Web Access Control
+
+ACL resources as defined by [Web Access Control](https://github.com/solid/web-access-control-spec) MUST be supported as a resource metadata type by Solid servers.
+
+ACL metadata resources are discoverable by the client via ```rel=acl```.
+
+An ACL metadata resource MUST be deleted when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
An ACL metadata resource MUST be deleted when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,176 @@
...
+
+A Solid server SHOULD sanity check ACL metadata resources upon creation or update to restrict invalid changes.
+
+### Resource Description
+
+Resource description is a general mechanism to provide descriptive metadata for a given resource. It MUST be supported as a resource metadata type by Solid servers.
+
+The Descriptive metadata resource for a given resource is discovered via ```rel=describedby```. Conversely, the resource being described by a Descriptive metadata resource is discovered via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A Descriptive metadata resource MUST be deleted when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Descriptive metadata resource MUST be deleted when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,176 @@
...
+
+Access or management of a Descriptive metadata resource by a given [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) is subject to the [modes of access](https://github.com/solid/web-access-control-spec#modes-of-access) granted per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+### Shape Validation
+
+Shape Validation ensures that any data changes in a given resource conform to an associated [SHACL](https://www.w3.org/TR/shacl/) or [ShEx](https://shex.io/shex-semantics/index.html) data shape. It MUST be supported as a resource metadata type by Solid servers.
+
+The Shape validation metadata resource for a given resource is discovered via ```rel=http://www.w3.org/ns/solid/terms#shape```. Conversely, the resource being described by a Shape validation metadata resource is discovered via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A Shape validation metadata resource MUST be deleted when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Shape validation metadata resource MUST be deleted when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,176 @@
...
+
+A Solid server SHOULD sanity check Shape validation metadata resources upon creation or update to restrict invalid changes.
+
+### Server Managed
+
+A Solid server stores information about a resource that clients can read but not change in Server Managed metadata. It MUST be supported as a resource metadata type by Solid servers.
+
+A Server Managed metadata resource is Discovered via ```rel=http://www.w3.org/ns/solid/terms#server```. Conversely, the resource being described by a Server Managed metadata resource is discovered via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A Server Managed metadata resource MUST be deleted when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Server Managed metadata resource MUST be deleted when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,176 @@
...
+
+To access a Server Managed metadata resource, an [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) MUST have [acl:Read](https://github.com/solid/web-access-control-spec#aclread) privileges per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+### Configuration
+
+Configuration metadata is used to store configurable parameters for a given resource. It MUST be supported as a resource metadata type by Solid servers.
+
+A configuration metadata resource is discovered via ```rel=http://www.w3.org/ns/solid/terms#configuration```. Conversely, the resource being described by a Configuration metadata resource is discovered via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A Configuration metadata resource MUST be deleted when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Configuration metadata resource MUST be deleted when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

justinwb:
Per the feedback in [meeting #10](https://github.com/solid/data-interoperability-panel/blob/master/meetings/10-20200113.md), I've pushed the following updates:
- [Client and server clarification](https://github.com/solid/data-interoperability-panel/pull/32/commits/885a8f848fc81790569703844fc90f7c8e320111) cc: @ericprud 
- [Moved draft resources into the readme](https://github.com/solid/data-interoperability-panel/pull/32/commits/52673a31a358d872eb30a3ad33ba29128895efdf)
- [Limited types to one metadata resource](https://github.com/solid/data-interoperability-panel/pull/32/commits/3b26f88077b2b7a08d32e260dffdd5d7dc437a56) cc: @elf-pavlik

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Web Access Control
+
+ACL resources as defined by [Web Access Control](https://github.com/solid/web-access-control-spec) MUST be supported as a resource metadata type by Solid servers.
+
+ACL metadata resources are discoverable by the client via ```rel=acl```.
+
+A given Solid resource MUST NOT be directly associated with more than one ACL metadata resource.
+
+An ACL metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
An ACL metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Resource Description
+
+Resource description is a general mechanism to provide descriptive metadata for a given resource. It MUST be supported as a resource metadata type by Solid servers.
+
+The Descriptive metadata resource for a given resource is discovered by the client via ```rel=describedby```. Conversely, the resource being described by a Descriptive metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A given Solid resource MUST NOT be directly associated with more than one Descriptive metadata resource.
+
+A Descriptive metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Descriptive metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Shape Validation
+
+Shape Validation ensures that any data changes in a given resource conform to an associated [SHACL](https://www.w3.org/TR/shacl/) or [ShEx](https://shex.io/shex-semantics/index.html) data shape. It MUST be supported as a resource metadata type by Solid servers.
+
+The Shape validation metadata resource for a given resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#shape```. Conversely, the resource being described by a Shape validation metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A given Solid resource MUST NOT be directly associated with more than one Descriptive metadata resource.
+
+A Shape validation metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Shape validation metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Server Managed
+
+A Solid server stores information about a resource that clients can read but not change in Server Managed metadata. It MUST be supported as a resource metadata type by Solid servers.
+
+A Server Managed metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#servermanaged```. Conversely, the resource being described by a Server Managed metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A given Solid resource MUST NOT be directly associated with more than one Server Managed metadata resource.
+
+A Server Managed metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Server Managed metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Configuration
+
+Configuration metadata is used to store configurable parameters for a given resource. It MUST be supported as a resource metadata type by Solid servers.
+
+A configuration metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#configuration```. Conversely, the resource being described by a Configuration metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A given Solid resource MUST NOT be directly associated with more than one Configuration metadata resource.
+
+A Configuration metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is also deleted and the Solid server is authoritative for both resources.
```
TallTed:
```suggestion
A Configuration metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```

resource-metadata/README.md
```diff
@@ -0,0 +1,54 @@
...
+- [How should we handle metadata of non-RDF sources](https://github.com/solid/solid-spec/issues/197)
+- [How to delete metadata file?](https://github.com/solid/solid-spec/issues/168)
+- [When a resource is deleted, should the acl be deleted too?](https://github.com/solid/solid/issues/251)
+- [When deleting a resources, should the corresponding ACL document be deleted?](https://github.com/solid/solid-spec/issues/187)
+- [Explain the ACL inheritance algorithm for clients](https://github.com/solid/specification/issues/106)
+
+Items/themes that are raised in the issues above and should be addressed in some form:
+
+- Reconfirm use of describedby as a way to get to meta information (#63)
+- Specify whether or not metadata should be handled as regular rdf resources (it should be) - probably should put this in the opening section.
```
TallTed:
```suggestion
- Specify that metadata should be handled as regular RDF resources - probably should put this in the opening section.
```

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
+# Resource Metadata Model
+
+## Background
```
csarven: High level suggestion: What people often conflate is metadata that's loosely coupled with a resource for the purpose of extending interaction possibilities or certain behaviours that comes out of it from the kind of data that extends the description of a resource. Making that distinction up front may help. The classic example has to do with how to deal with providing a human-readable label for a container (or from LDP, updating "minimal-container triples"). That use case or anything that's about self-describing resources is outside the scope of "resource metadata" (as I see it and hopefully I'm not off the mark with what the panel intends to do). What's described here is really about the ones that's highlighted: access control, annotation, shapes, server-managed, configuration, and possibly others defined elsewhere. This is important because we still want the primary resource to be useful in its own right without forcing the client to navigate to the the related meta resource (eg. just to get a hold of container's label) - that's besides annotating non-RDF Sources. That also opens up the possibility to progressive enhance so to speak, and if so desired.

It may help to just change the order of information in the Background.

I sense some redundancy with the two example blocks. Perhaps that can be a single list that's tied to use cases eg. at least one for each metadata type.

The section should be condensed as much as possible. It could just be an "Overview" - let's see..

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+Some examples include:
+
+- An ACL resource controls how a Solid server makes authorization decisions for a container and any child resources.
+- A shape constraint resource may limit the RDF structures that can be added to a resource or container.
+- A “configuration” resource may control how a resource is versioned, or which indexes are exposed for it.
+
+Access to different categories of metadata may require different levels of privilege, which must be specified as part of the definition for a given metadata type.
+
+## Metadata Discovery
```
csarven: To give an overview, I think text and table along the the following lines can help (with some inspiration from the fragment specification table under https://www.w3.org/TR/annotation-model/#model-18 ):

"
The following are some of the resource metadata types and the relation URIs that are used to discover the resource metadata. Other metadata types and relations may also be used.

Type | Relation | Description
--- | --- | ---
Access control | `acl` or `http://www.w3.org/ns/solid/terms#acl` | [[WAC](http://solid.github.io/web-access-control-spec/)]
Resource description  | `describedby` or `https://www.w3.org/ns/iana/link-relations/relation#describedby` | ...
Shape | `http://www.w3.org/ns/solid/terms#shape` | ...
Server managed | `http://www.w3.org/ns/solid/terms#servermanaged` | ...
Configuration | `http://www.w3.org/ns/solid/terms#configuration` | ...

"

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+Given the URL of a resource, a client can discover the metadata resources by issuing a `GET` (or `HEAD`) request and inspecting the `Link` headers in the response. The [`rel={attribute}`](https://tools.ietf.org/html/rfc8288) will define the relationship to the target URL.
+
+For any defined metadata type available for a given resource, all representations of that resource MUST include a Link header pointing to the location of each metadata resource. For example, as defined by the Solid [Web Access Control specification](https://github.com/solid/web-access-control-spec), a client can use this mechanism to discover the location of an ACL resource:
+
+Link: <https://example.com/resource?acl>; rel="acl"
+
+Metadata discovered through a Link header for a given resource is considered to be *directly associated* with that resource.
+
+### Discovery of Annotated Resource
```
csarven: Consider information table like in https://github.com/solid/data-interoperability-panel/pull/32/files#r371016439 or have a single table combining relations from both sides.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+Some examples include:
+
+- An ACL resource controls how a Solid server makes authorization decisions for a container and any child resources.
+- A shape constraint resource may limit the RDF structures that can be added to a resource or container.
+- A “configuration” resource may control how a resource is versioned, or which indexes are exposed for it.
+
+Access to different categories of metadata may require different levels of privilege, which must be specified as part of the definition for a given metadata type.
+
+## Metadata Discovery
+Given the URL of a resource, a client can discover the metadata resources by issuing a `GET` (or `HEAD`) request and inspecting the `Link` headers in the response. The [`rel={attribute}`](https://tools.ietf.org/html/rfc8288) will define the relationship to the target URL.
```
csarven: Is it restricted to `GET` and `HEAD`? How about `OPTIONS`?

Should `rel={attribute}` be `rel={relationtype}` ?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
+# Resource Metadata Model
+
+## Background
+
+*This section is non-normative*
+
+This document introduces a mechanism for linking to metadata about resources in the Solid Ecosystem using HTTP Link headers. Examples of this linking mechanism include:
+
+- A binary JPEG image with a Link header, pointing to a description that has an RDF representation.
+- An LDP container with a Link header, pointing to an ACL resource that governs access controls for the contained resources.
```
csarven: I'd suggest not to mention LDP container but just go with "container" because the Solid container (`/`) is not strictly defined as an LDP container. So, just `s/An LDP/A` would suffice.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+- A “configuration” resource may control how a resource is versioned, or which indexes are exposed for it.
+
+Access to different categories of metadata may require different levels of privilege, which must be specified as part of the definition for a given metadata type.
+
+## Metadata Discovery
+Given the URL of a resource, a client can discover the metadata resources by issuing a `GET` (or `HEAD`) request and inspecting the `Link` headers in the response. The [`rel={attribute}`](https://tools.ietf.org/html/rfc8288) will define the relationship to the target URL.
+
+For any defined metadata type available for a given resource, all representations of that resource MUST include a Link header pointing to the location of each metadata resource. For example, as defined by the Solid [Web Access Control specification](https://github.com/solid/web-access-control-spec), a client can use this mechanism to discover the location of an ACL resource:
+
+Link: <https://example.com/resource?acl>; rel="acl"
```
csarven: I acknowledge the intention with the example use of `?acl` but to minimise any confusion eg. is the ACL resource somehow a query on a resource (or even if a Solid server is capable of taking query params), as well as to emphasise the notion that the ACL resource is really a distinct resource, use something like `https://example.org/resource.acl`

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+Metadata discovered through a Link header for a given resource is considered to be *directly associated* with that resource.
+
+### Discovery of Annotated Resource
+
+Certain metadata resource types MAY require the Solid server to include a link back to the annotated resource using an appropriate link relation. For example, LDP defines a bidirectional discovery mechanism for RDF descriptions of NonRDF resources, via Link headers:
```
csarven: "MAY require"?

No need to mention LDP behaviour.

`s/NonRDF/non-RDF`

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+Link: <https://example.com/binary?description>; rel="description"
```
csarven: Should `rel="description"` be `rel="describes"`?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+along with:
+
+Link: <https://example.com/binary>; rel="describedby"
+
+For cases where link relations are not defined by IANA, a URL can be used. For example:
```
csarven: I think the first sentence would suffice for this block by just referring to RFC Web Linking ( https://tools.ietf.org/html/rfc8288#section-2.1 ) on allowed relation type ie. registered terms at IANA or URI. This information is applicable to both Metadata Discovery and Discovery of Annotated Resource sections. I think it should come before those two sections... either a unique section describing common requirements or maybe even move it to Overview.

If you really want URL, as opposed to a URI, make sure that's well-defined somewhere. On a related note, there is discussion in solid/specification about dereferenceable and machine-readable constraints (eg. target of `ldp:constrainedBy`). I would put these things in the same category even though the former is about the vocab and the latter is some resource.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+and:
+
+Link: <https://example.com/resource>;
+        rel="https://example.org/ns#isMetadataOf"
+
+String-based link relations, such as in the examples above, must be registered with IANA. But it is also possible to use custom relation types by using a full IRI. The [Linked Data Platform](https://www.w3.org/TR/ldp/), the [Linked Data Notifications](https://www.w3.org/TR/ldn/), and the [Web Annotation Protocol](http://www.w3.org/TR/annotation-protocol/) specifications make use of full IRIs in the `rel` attribute.
+
+## Metadata Characteristics
+
+A given resource MAY Link to metadata on a different server.
```
csarven: How? Just by client referring to it when the resource is created? Or is this something only the server may establish?

If and how can the URI be updated?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+Metadata resources on the same Solid server MUST adhere to the same interaction model used by standard Solid resources.
```
csarven: This is a bit overloaded I think.. or just unclear. Is it just that metadata resources are like every other resource? Is the key point about the resource lifecycle?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+ACL resources as defined by [Web Access Control](https://github.com/solid/web-access-control-spec) MUST be supported as a resource metadata type by Solid servers.
+
+ACL metadata resources are discoverable by the client via ```rel=acl```.
+
+A given Solid resource MUST NOT be directly associated with more than one ACL metadata resource.
+
+An ACL metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
+
+To access or manage an ACL metadata resource, an [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) MUST have [acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol) privileges per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
```
csarven: I don't think this information needs to be said (and may even go stale if changed at source; WAC). Perhaps something like this (which similar to the ones in the other metadata categories) move into an earlier section about all categories. Like the example with deleting a resource also deletes its metadata?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Web Access Control
+
+ACL resources as defined by [Web Access Control](https://github.com/solid/web-access-control-spec) MUST be supported as a resource metadata type by Solid servers.
+
+ACL metadata resources are discoverable by the client via ```rel=acl```.
+
+A given Solid resource MUST NOT be directly associated with more than one ACL metadata resource.
+
+An ACL metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
```
csarven: Move delete information that applies to all categories to parent section.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+To access or manage an ACL metadata resource, an [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) MUST have [acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol) privileges per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+A Solid server SHOULD sanity check ACL metadata resources upon creation or update to restrict invalid changes.
```
csarven: Nice but perhaps underspecified. Is this supposed to be about having a shape for ACL? Are there cases where a server may let an invalid change stay?

Is this actually about resource metadata or resources in general? If latter, it should refer to some common behaviour.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Resource Description
+
+Resource description is a general mechanism to provide descriptive metadata for a given resource. It MUST be supported as a resource metadata type by Solid servers.
+
+The Descriptive metadata resource for a given resource is discovered by the client via ```rel=describedby```. Conversely, the resource being described by a Descriptive metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
```
csarven: What's `solid:resource` supposed to be? Aside: I don't like the name. What happened to `describes`?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+String-based link relations, such as in the examples above, must be registered with IANA. But it is also possible to use custom relation types by using a full IRI. The [Linked Data Platform](https://www.w3.org/TR/ldp/), the [Linked Data Notifications](https://www.w3.org/TR/ldn/), and the [Web Annotation Protocol](http://www.w3.org/TR/annotation-protocol/) specifications make use of full IRIs in the `rel` attribute.
+
+## Metadata Characteristics
+
+A given resource MAY Link to metadata on a different server.
+
+Metadata resources on the same Solid server MUST adhere to the same interaction model used by standard Solid resources.
+
+## Reserved Metadata Types
```
csarven: What's the actual process to refer and create resource metadata?

Are resource metadata listed in containment? https://github.com/solid/specification/issues/116

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+The Descriptive metadata resource for a given resource is discovered by the client via ```rel=describedby```. Conversely, the resource being described by a Descriptive metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A given Solid resource MUST NOT be directly associated with more than one Descriptive metadata resource.
+
+A Descriptive metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
+
+Access or management of a Descriptive metadata resource by a given [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) is subject to the [modes of access](https://github.com/solid/web-access-control-spec#modes-of-access) granted per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+### Shape Validation
```
csarven: [What shapes the shapes?  Okay, I'll see myself out.]

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+To access or manage a Shape validation metadata resource, an [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) MUST have [acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol) privileges per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+A Solid server SHOULD sanity check Shape validation metadata resources upon creation or update to restrict invalid changes.
+
+### Server Managed
+
+A Solid server stores information about a resource that clients can read but not change in Server Managed metadata. It MUST be supported as a resource metadata type by Solid servers.
+
+A Server Managed metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#servermanaged```. Conversely, the resource being described by a Server Managed metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
```
csarven: Minor: Can we use a one word term instead of "servermanaged"?

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+A Shape validation metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
+
+To access or manage a Shape validation metadata resource, an [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) MUST have [acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol) privileges per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+A Solid server SHOULD sanity check Shape validation metadata resources upon creation or update to restrict invalid changes.
+
+### Server Managed
+
+A Solid server stores information about a resource that clients can read but not change in Server Managed metadata. It MUST be supported as a resource metadata type by Solid servers.
```
csarven: Perhaps give more detail about "server managed" or tie it to use cases.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+A given Solid resource MUST NOT be directly associated with more than one Server Managed metadata resource.
+
+A Server Managed metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
+
+To access a Server Managed metadata resource, an [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) MUST have [acl:Read](https://github.com/solid/web-access-control-spec#aclread) privileges per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+### Configuration
+
+Configuration metadata is used to store configurable parameters for a given resource. It MUST be supported as a resource metadata type by Solid servers.
```
csarven: Perhaps give more detail about "configuration" or tie it to use cases.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+A configuration metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#configuration```. Conversely, the resource being described by a Configuration metadata resource is discovered by the client via ```rel=http://www.w3.org/ns/solid/terms#resource```.
+
+A given Solid resource MUST NOT be directly associated with more than one Configuration metadata resource.
+
+A Configuration metadata resource MUST be deleted by the Solid server when the resource it is directly associated with is deleted and the Solid server is authoritative for both resources.
+
+To access or manage a Configuration metadata resource, an [acl:agent](https://github.com/solid/web-access-control-spec#describing-agents) MUST have [acl:Control](https://github.com/solid/web-access-control-spec#aclcontrol) privileges per the [ACL inheritance algorithm](https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm) on the resource directly associated with it.
+
+## Non-Reserved Types
```
csarven: I'd suggest to express this in terms of "Extensions" and mention that it needs to follow the base or global criteria as the other metadata types.

resource-metadata/hypothesis.md
```diff
@@ -0,0 +1,139 @@
...
+
+A Solid server may support other metadata types.
+
+## Implementation Patterns
```
csarven: I don't know if you plan to extend this section but may want to tuck it under a section about common behaviour early on.

csarven:
I propose to include the following criteria for clients for discovering resources and metadata resources (adapted from https://www.w3.org/TR/ldn/#discovery ):
"
Clients do the following to discover resources and their metadata:

* make an HTTP `HEAD` or `GET` request on the target URL, and use the `Link` header with a `rel` value of a metadata type.
* make an HTTP `GET` request on the target URL to retrieve an RDF representation [RDF 1.1], whose encoded RDF graph contains a relation of metadata type.

These may be carried out in either order, but if the first fails to result in discovering the metadata resource or the resource the second MUST be tried.
"

resource-metadata/hypothesis.md
elf-pavlik: @ericprud made PR with shapes for ACL resources https://github.com/solid/web-access-control-spec/pull/77/files I think we could reference them instead of talking in terms of "sanity check".

csarven:
If the role of ldp:constrainedBy ( https://github.com/solid/specification/issues/44 ) is in scope, it should be stated.

resource-metadata/hypothesis.md
justinwb: In the current definition this is something only the server may establish. The purpose of this was to point out that a server could be configured to link to metadata resources somewhere else.

resource-metadata/hypothesis.md
justinwb: Yes - the point is to say that if this resource is a solid resource, it abides by the same interaction model as any other solid resource. We say "the same solid server" because the server may not know that the link to a resource somewhere else is a resource on a solid server.

resource-metadata/hypothesis.md
justinwb: We say now in the Metadata Characteristics section that permissions will be defined in the definition of a given type. The ACL permissions differ by metadata type, so it wouldn't be appropriate to say more than that. The delete behavior is consistent across the currently defined types. A new type could have reason to be handled differently, but delete would be okay to move into a general section given the current set. We decided early on that we'd overload the individual types and put less in the general description, which is why we include it in each one.

resource-metadata/hypothesis.md
justinwb: The server through configuration determines where metadata resources live and provides that in link headers. Assuming a given metadata resource is on the same solid server, it should be listed in containment in that location, and follow the same interaction model as other solid resources, with the exception of any special handling due a resource of a given metadata type.

resource-metadata/hypothesis.md
justinwb: Suggestions welcome!! 😄

resource-metadata/hypothesis.md
justinwb: Incorporated suggestions and condensed overview.

resource-metadata/hypothesis.md
justinwb: Added OPTIONS and changed to ```rel={relation-type}```

resource-metadata/hypothesis.md
justinwb: Removed LDP

resource-metadata/hypothesis.md
justinwb: Adjusted language for clarity.

resource-metadata/hypothesis.md
justinwb: Per suggestion, pointed to RDF8288 section on allowed types, and removed the latter section that described what is covered in the linked reference.

resource-metadata/hypothesis.md
justinwb: I think shape is part of the solution, but not all. There is some permission logic that a shape can't cover on its own.

resource-metadata/hypothesis.md
justinwb: shortened the term to ```rel=http://www.w3.org/ns/solid/terms#managed``` 😐 😄

justinwb:
> These may be carried out in either order, but if the first fails to result in discovering the metadata resource or the resource the second MUST be tried.

@csarven MUST feels a bit strong here. Is SHOULD good enough?

justinwb:
> If the role of ldp:constrainedBy ( [solid/specification#44](https://github.com/solid/specification/issues/44) ) is in scope, it should be stated.

@csarven I could see a role for ldp:constrainedBy in response when something fails Access Control or Shape Validation (or something similar in the future).

csarven:
>>These may be carried out in either order, but if the first fails to result in discovering the metadata resource or the resource the second MUST be tried.

>@csarven MUST feels a bit strong here. Is SHOULD good enough?

MUST is required for interop in that it guarantees discovery by trying both possibilities. The key point about the requirement is setting an implementation's capability ie. checking both places one after another for any resource. Whether they want to or not for a particular target resource is at their discretion. For example, the LDN spec notes targets with fragment identifiers and that there is no point in checking the Link header to discover the inbox because the relation can only be found in RDF body. Similarly, for a non-RDF target resource eg. binary image, only the Link header needs to be checked.

csarven:
>>If the role of ldp:constrainedBy ( solid/specification#44 ) is in scope, it should be stated.

>@csarven I could see a role for ldp:constrainedBy in response when something fails Access Control or Shape Validation (or something similar in the future).

That's essentially what the LDP spec recommends. I was hoping for something more concrete re: https://gitter.im/solid/data-interoperability-panel?at=5e3abdde594a0517c26ac1fb given the alternative take with solid:shape. Basically addressing issue 44 (perhaps specific questions in https://github.com/solid/specification/issues/44#issuecomment-540991655 can shed light?)

justinwb:
> > > If the role of ldp:constrainedBy ( [solid/specification#44](https://github.com/solid/specification/issues/44) ) is in scope, it should be stated.
> 
> > @csarven I could see a role for ldp:constrainedBy in response when something fails Access Control or Shape Validation (or something similar in the future).
> 
> That's essentially what the LDP spec recommends. I was hoping for something more concrete re: https://gitter.im/solid/data-interoperability-panel?at=5e3abdde594a0517c26ac1fb given the alternative take with solid:shape. Basically addressing issue 44 (perhaps specific questions in [solid/specification#44 (comment)](https://github.com/solid/specification/issues/44#issuecomment-540991655) can shed light?)

@csarven Added a comment on solid/specification#44. For starters I will update the text to illustrate the use of ldp:constrainedBy in response to a request that violates each type of constraint.

justinwb:
> @csarven Added a comment on [solid/specification#44](https://github.com/solid/specification/issues/44). For starters I will update the text to illustrate the use of ldp:constrainedBy in response to a request that violates each type of constraint.

As I sit down to do this - it occurs to me that explaining how ldp:constrainedBy is incorporated into responses based on a given constraint probably belongs in a separate section. Either it should go in a section specific to error handling, and/or it should go into a section specific to how constraints are defined and/or evaluated. The section for resource metadata focuses on how you associate that there is a constraint associated with a resource, but doesn't go into detail on how that constraint be defined and/or evaluated.

resource-metadata/hypothesis.md
csarven: I'm not sure if there is anything prescribed here that can be tested for required interop besides that a server should be capable of including Link headers for metadata relations for a resource and that the target of the relation can be under any authority. Everything else seems to be out of scope.

The prescription for server's capability can actually be derived from discovery mechanism: https://github.com/solid/data-interoperability-panel/pull/32#issuecomment-578549345

Would this clarify:

"Servers that want to associate a metadata resource to a resource MUST either use the Link header with a rel value of a metadata type or asset the relation in the encoded RDF graph of resource's representation."

justinwb:
At this point resolutions have been committed based on feedback, and this text is now being moved into bikeshed format to be submitted as a candidate proposal to the https://github.com/solid/specification. No need to maintain this pull for collaborative editing.

PR #41, opened 2020-05-18
moving UCR from auth panel

elf-pavlik:
This suppose to provide starting point for Use Cases and Requirements document. Current content comes directly from https://github.com/solid/authorization-and-access-control-panel/pull/66/files

**EDIT** [rendered HTML preview](https://gistpreview.github.io/?408e2909670b623f9341cdf83a71a95f)

### DIFF
--- a/use-cases-and-requirements.bs
+++ b/use-cases-and-requirements.bs
@@ -1,0 +1,219 @@
+<pre class='metadata'>
+Title: Solid Use Cases and Requirements
+Shortname: your-spec
+Level: 1
+Status: w3c/ED
+Group: solid-cg
+URL: http://example.com/url-this-spec-will-live-at
+Editor: elf Pavlik
+Abstract: Use Cases and Requirements for Solid ecosystem.
+</pre>
+
+Introduction {#intro}
+=====================
+
+Issue: Needs explaination how proposals should clearly reference requirements which they satisfy. As well as requirements they do not satisfy due to some limitations.
+
+Requirements {#requirements}
+============================
+
+## The system is not abusable
+
+<div class="assertion" heading="Use cases">
+* [[#evil-fitness-tracker]]
+</div>
+
+## An app can request access to a specific resource
+
+<div class="assertion" heading="Use cases">
+* [[#simple-game]]
+* [[#rdf-editor]]
+* [[#pod-administration]]
+</div>
+
+## An app can request access to a specific type of data without knowing the structure of resources on a Pod
+
+<div class="assertion" heading="Use cases">
+* [[#chat-with-friends]]
+* [[#discussion-boards]]
+* [[#photos-viewer]]
+* [[#photos-processing]]
+</div>
+
+## Access requests can be sent when the resource owner is not present to be approved once the user is present
+
+## Apps can request the ability to write a specific type of data and will be told where it should write it
+
+<div class="assertion" heading="Use cases">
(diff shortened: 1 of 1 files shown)


csarven:
Great kick off! I suggest changing the order; Use Cases normally precede Requirements.

If you can connect the UCs to the User Stories at https://github.com/solid/user-stories (see issues), that'd be very helpful. We also need some kind of feedback from the community (survey).

See the first comment in https://github.com/solid/specification/issues/9 for general direction and also links to example UCRs from W3C TR.

justinwb:
@elf-pavlik had a [discussion](https://www.w3.org/community/solid/wiki/Meetings#Use_Cases) with @csarven on friday about putting these directly into the  [solid/specification](https://github.com/solid/specification) repo. For myself I think this is a good idea. I think we can tune it up a bit in here (get the structure right, etc), and then submit it into the specification repo as a proposal. IMO we should try to get that done quickly in here, and then operate on it directly under the solid/specification repo moving forward

use-cases-and-requirements.bs
```diff
@@ -0,0 +1,219 @@
...
+Group: solid-cg
+URL: http://example.com/url-this-spec-will-live-at
+Editor: elf Pavlik
+Abstract: Use Cases and Requirements for Solid ecosystem.
+</pre>
+
+Introduction {#intro}
+=====================
+
+Issue: Needs explaination how proposals should clearly reference requirements which they satisfy. As well as requirements they do not satisfy due to some limitations.
```
justinwb, line 15: I think it's also important to provide some expectations around use case scope. We should focus mostly on lean and easy to understand use cases and requirement that don't drift too far away from practical use, so they can be as informative and referenceable as possible.

use-cases-and-requirements.bs
```diff
@@ -0,0 +1,219 @@
...
+
+Requirements {#requirements}
+============================
+
```
justinwb, line 19: I feel like requirements should be categorized (e.g. authentication, authorization, identity, validation, resource access, security), but use cases should specifically not be, since good ones should span categories.

use-cases-and-requirements.bs
```diff
@@ -0,0 +1,219 @@
...
+## Data should have different levels of requirements for user's conciousness in consent
+
+## The authorization data is easily cachable
+
+## Information about Clients (apps) to which users have granted access and what specific access they have delegated should not be made available to Resource Servers that do not enforce relevant access restrictions on those Clients.
+
+
+Use Cases {#use-cases}
+======================
+
```
justinwb, line 74: It feels like the use cases should be organized more based on the unique pattern that they're demonstrating, so they can be found and navigated more intuitively.

elf-pavlik:
Instead of trying to address all the ideas in this PR I would prefer to merge it and invite everyone to submit following up PRs with amendments. In case anyone sees problem but doesn't feel ready to submit PR we can open an issue and have preliminary discussion there.

> I think we can tune it up a bit in here (get the structure right, etc), and then submit it into the specification repo as a proposal.

SGTM let's iterate via multiple PRs

csarven:
I've no objection to moving any changes to a different PR but I'm not sure if a review here serves much purpose. Go for it I guess.

justinwb:
Merging this in - will iterate in successive PRs to incorporate some of the feedback left in this one.

PR #42, opened 2020-06-01
Ecosystem - introduction and application registration

justinwb:
First of what will be a multi-part submission for an interoperable client ecosystem. This pull request focuses on providing some introductory text, use case, definitions, environment, and an approach for application registration for further panel iteration.

### DIFF
--- a/proposals/ecosystem.bs
+++ b/proposals/ecosystem.bs
@@ -134,6 +134,10 @@ thing, data model, abstract entity, etc.) with a DID document,
 equivalent to an [=identity profile document=], to allow trustable
 interactions with that subject. [[DID]]
 
+An <dfn>identity provider</dfn> is...
+
+A <dfn>bearer token</dfn> is...
+
 An <dfn>ApplicationID</dfn> is a web resource at an HTTP URI uniquely associated
 with a given [=application=]. For [=server-side applications=], it may also
 serve as an [=identity=]. An [=application profile document=] can be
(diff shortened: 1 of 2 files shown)


proposals/ecosystem.bs
```diff
@@ -0,0 +1,1254 @@
...
+application registry that stores application registrations in
+`<https://alice.pod.example/applications/>`.
+
+<b>Excerpt from [[#alice-appregset]] Document</b>
+<pre highlight="turtle">
+@prefix solid-ecosystem: &lt;http://www.w3.org/ns/solid/ecosystem#&gt; .
+
+<#set>
+  a solid-ecosystem:registrySet ;
+  solid-ecosystem:registryType solid-ecosystem:ApplicationRegistry ;
```
joshdcollins, line 321: I believe we'd want :ApplicationRegistry to be :applicationRegistry -- we seem to be inconsistent in the capitalization.

RubenVerborgh: `ApplicationRegistry` seems to be an instance; they are usually capitalized.

RubenVerborgh: Reading this made me wonder why they are not RDF types though.

joshdcollins: We had some debate on this -- landed on the case that (as of now) there is no difference between an application registry and a data registry, aside from the value of the :registryType predicate.

This :registryType is validated using the ShEx [here](https://github.com/solid/data-interoperability-panel/pull/42/files#diff-a2a3ce25ca75fa5d1aa87cbaa8c2602dR1034) to ensure a registry set for applications (ApplicationRegistry) does not have members of type DataRegistry.

acoburn: > Reading this made me wonder why they are not RDF types though.

RDF types would seem to make sense here. E.g.:

```turtle
ecosystem:Registry a rdfs:Class .
ecosystem:ApplicationRegistry a ecosystem:Registry .

<#set> a ecosystem:ApplicationRegistry .
```

justinwb: Fully expect and welcome tons of feedback on the data model. We put more of a premium on getting something together and presented soon even if it wasn't pristine yet so that we can all iterate on it together. We're reusing most of these patterns so the faster we can nail them down the cleaner the next submissions will be 🙂

justinwb: Created [this issue](https://github.com/solid/data-interoperability-panel/issues/45) to continue this conversation. We're going to merge and iterate on this through subsequent pulls and issues.

PR #59, opened 2020-07-31
Data authorization

justinwb:
Introduces Data Authorization workflows, with Access Needs, Access Grants, Access Receipts, and Remote Data Registration.


proposals/ecosystem/index.bs
```diff
@@ -1610,9 +1619,306 @@ Each new [=Registered Data Instance=] must be linked to the
...
+* [eco:hasSeries](ecosystem.ttl#hasSeries) - links to one
+    [=Access Decorator Series=].
+
+<figure id="fig-access-need-decorator-index">
+  <figcaption>[=Access Decorator Index=] referenced in the
+  [previous figure](#fig-access-need-group)</figcaption>
+  <pre highlight="turtle">
+  <#notebook-access-decorator-index>
+      a eco:AccessDecoratorIndex ;
+      eco:defaultLanguage "en" ;
```
joshdcollins, line 1808: I believe we want to use xsd:lang or xsd:language

proposals/ecosystem/index.bs
```diff
@@ -1805,7 +1805,7 @@ can be invoked with the following properties:
   <pre highlight="turtle">
   <#notebook-access-decorator-index>
       a eco:AccessDecoratorIndex ;
-      eco:defaultLanguage "en" ;
+      eco:defaultLanguage "en"^^xsd:lang ;
```
csarven, line 1808: `rdf:langString`

proposals/ecosystem/index.bs
```diff
@@ -1825,7 +1825,7 @@ can be invoked with the following properties:
   <pre highlight="turtle">
   <#series-en>
       a eco:AccessDecoratorSeries ;
-      eco:usesLanguage "en" ;
+      eco:usesLanguage "en"^^xsd:lang ;
```
csarven, line 1828: `rdf:langString`

proposals/ecosystem/index.bs
csarven: I'm not sure what's intended actually. xsd:lang doesn't exist but there is xml:lang. Is it supposed to be rdf:langRange instead of eco:defaultLanguage?

ericprud: I think this was meant to be xsd:language. They both ([`xml:lang`](https://www.w3.org/TR/2008/REC-xml-20081126/#sec-lang-tag), [`xsd:langauge`](https://www.w3.org/TR/xmlschema-2/#language)) expect BCP47 language codes (noting that [`BCP47` obsoletes `RFC3066`](https://www.w3.org/International/articles/language-tags/)) but `xsd` is used all time in RDF.

justinwb: Updated to `xsd:language`

proposals/ecosystem/index.bs
```diff
@@ -1610,9 +1627,438 @@ Each new [=Registered Data Instance=] must be linked to the
...
+  </pre>
+</figure>
+
+An [=Access Need Group Decorator=] is represented by the class
+[eco:AccessNeedGroupDecorator](ecosystem.ttl#AccessNeedGroupDecorator),
+and can be invoked with the following properies:
+
+* [eco:hasAccessNeedGroup](ecosystem.ttl#hasAccessNeedGroup) - Links to
+    the [=Access Need Group=] being described
+* <a href="http://www.w3.org/2004/02/skos/core#prefLabel" class="vocab">
```
dmitrizagidulin, line 1728: Any particular reason to use skos `prefLabel` here, and not `rdfs:label` as is used elsewhere in solid spec?

dmitrizagidulin:
Very interesting direction! At least on first reading, I'm finding the multi-part AccessDecorator(Index/Series/etc) system to be really hard to understand. But maybe it'll get clearer on multiple readings.

proposals/ecosystem/index.bs
```diff
@@ -1610,9 +1627,438 @@ Each new [=Registered Data Instance=] must be linked to the
...
+* [eco:accessMode](ecosystem.ttl#accessMode) - Links to one or more
+    [=access modes=]: `acl:Read`, `acl:Write`,
+    `acl:Append`, `acl:Control`
+
+<figure id="fig-access-need">
+  <figcaption>An [=Access Need=] for the `<note:notebook-tree>`</figcaption>
+  <pre highlight="turtle">
+  <#notebook>
+      a eco:AccessNeed ;
+      eco:inAccessNeedGroup &lt;#notebook-collaboration-group&gt; ;
```
elf-pavlik, line 1771: Why not to just reuse `eco:requestsAccess` in reverse direction?

proposals/ecosystem/ecosystem.shex
```diff
@@ -87,3 +87,44 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
...
+  eco:hasAccessDecoratorIndex IRI ;
+  eco:needGroupLabel xsd:string ;
+  eco:needGroupExplanation xsd:string
+}
+
+
+# Validate an instance of the eco:AccessNeed class
+<#PreparedAccessNeedShape> {
+  a [ eco:AccessNeed ] ;
+  eco:inAccessNeedGroup IRI+ ;
```
elf-pavlik, line 119: this very likely could use `^eco:requestsAccess IRI ;` and not rely on defining inverse property
https://shex.io/shex-primer/#inverse-properties

ping @ericprud

justinwb:
> Very interesting direction! At least on first reading, I'm finding the multi-part AccessDecorator(Index/Series/etc) system to be really hard to understand. But maybe it'll get clearer on multiple readings.

Will be adding text to explain that, but essentially the decorators allow us to add labels and descriptions to data without tying them directly to it. This allows us to have access needs that can be described in different languages, or (depending on the context) different ways

proposals/ecosystem/ecosystem.shex
ericprud: I'm certainly in favor of having using only one direction or the other. WRT which, I guess I like the arcs-in (`<<group>> eco:includesNeed <<need>>`) better 'cause we don't have to invent URLs for all of the needs.

I used `includesNeed` rather than `requestsAccess` 'cause we currently have a distinction between root rules and overrides. I want to play with a few use cases to figure out whether they can be the same thing so in the mean time, `includesNeed` preserves the current expressivity.

proposals/ecosystem/index.bs
ericprud: We currently use `skos:narrower` to create a UI hierarchy. Some generic SKOS tools might be able to do something useful with that, plus the multi-lang conventions for `skos:prefLabel` are better-established than for `rdfs:label`.

ericprud:
> > Very interesting direction! At least on first reading, I'm finding the multi-part AccessDecorator(Index/Series/etc) system to be really hard to understand. But maybe it'll get clearer on multiple readings.
> 
> Will be adding text to explain that, but essentially the decorators allow us to add labels and descriptions to data without tying them directly to it. This allows us to have access needs that can be described in different languages, or (depending on the context) different ways

Maybe this would be improved by a leading example UI which associates features (text comes from the AppDecorator) with their required resources (text comes from the ShapeTreeDecorator).

(i miss email threads, quoting is a noisy substitute)

proposals/ecosystem/index.bs
```diff
@@ -5175,12 +5175,533 @@ Remote Data Registry {#remote}
...
+</figure>
+
+<figure id="fig-remote-registration">
+  <figcaption>An [=Remote Data Registration=] at
+  https://bob.pod.example/remote/f0ebbd29-3a5a-4770-af29-cdb7693066b4/#registration
+  </figcaption>
+  <pre highlight="turtle">
+  <#registration>
+    a eco:RemoteDataRegistration ;
+    eco:registeredBy https://bob.pod.example/profile/id#me ;
```
elf-pavlik: this and next snippet start missing < > around IRIs,

joshdcollins (approved):

PR #62, opened 2020-09-09
initial data registrations

### LINKED ISSUE #61: Add examples of Data Registries for Projects and Tasks
I think we can add some Data Registries for Projects and Tasks. I think `acme.example` would be most interesting one to start with. We have 3 projects there with different people having different kinds of access to them. We also have this requirement: "moving task between projects should not cause URL to change".

How do you see Data Registries applying here? Would `acme.example` only have one registry for Projects and another one for Tasks?
I guess for access control on Tasks we would rely on [Conditional access by relationship](https://solid.github.io/authorization-panel/wac-ucr/#conditional-relationship). Given that [Task shape](https://solid.github.io/data-interoperability-panel/primer/#tasks) mandates use of `ex:inProject` predicate.

elf-pavlik:
closes #61 

what else do you think this PR should include?


elf-pavlik:
@dmitrizagidulin @jaxoncreed It would be great to get some feedback from both of you on the `solidshapes.example`. It supposed to act as common repository of shapes and shape trees. I recall that both of you work on crating something in those lines.

justinwb:
> @dmitrizagidulin @jaxoncreed It would be great to get some feedback from both of you on the `solidshapes.example`. It supposed to act as common repository of shapes and shape trees. I recall that both of you work on crating something in those lines.

@elf-pavlik we've secured the shapes.pub and shapetrees.pub domains that can be safely used if we need real domains. if we're talking about examples that don't need to resolve then i think we can use shapes.example and shapetrees.example

justinwb (approved):

PR #65, opened 2020-10-05
Refactor app registration

justinwb:
Update app registration to use cleaned up data model, resource hierarchy, operations, permission model section format.

### DIFF
--- a/proposals/specification/definitions.bs
+++ b/proposals/specification/definitions.bs
@@ -46,12 +46,6 @@ thing, data model, abstract entity, etc.) with a DID document,
 equivalent to an [=identity profile document=], to allow trustable
 interactions with that subject. [[DID]]
 
-An <dfn>identity provider</dfn> furnishes an [=agent=] with authenticated
-credentials that prove agent is who it says it is, across the web.
-
-A <dfn>token</dfn> is used to make authenticated requests to the [=Solid=]
-server for a given [=authenticated agent=].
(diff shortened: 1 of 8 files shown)


proposals/specification/definitions.bs
```diff
@@ -106,33 +97,6 @@ related resources as a single grouping, and apply that to a range of
...
-An <dfn>Application Registration</dfn> provides the [=Agent=]
-with a place to maintain metadata, state, preferences, and
-other application-specific data associated with a given [=Application=] they
-have elected to use. Represented by the
-[interop:ApplicationRegistration](interop.ttl#ApplicationRegistration) class.
-Validated by the
-[interops:ApplicationRegistrationShape](interop.shex#ApplicationRegistrationShape).
-Associated with resources by the
-[interopt:ApplicationRegistrationTree](interop.tree#ApplicationRegistrationTree).
-
```
ericprud, line 135: Did the above three defns (Application Registry Set, Application Registry, Application Registration) migrate to another doc or were they judged superfluous?

ericprud (approved):

proposals/specification/definitions.bs
justinwb: They're still in there - but I've been moving definitions inline to the actual place where they're introduced (in this case in the application registration section), rather than at the end of the file where there's less context :)

PR #68, opened 2020-11-02
minimal setup for netlify

elf-pavlik:
I think we will need something like that in netlify app:

image


proposals/index.html
```diff
@@ -0,0 +1,17 @@
...
+  <meta charset="utf-8">
+
+  <title>Solid Data Interoperability Panel: Proposals</title>
+</head>
+
+<body>
+  <ul>
+    <li><a href="problems-and-goals/">Problems and Goals for Interoperability in Solid</a></li>
+    <li><a href="specification/">Solid Application Interoperability</a></li>
+    <li><a href="primer/">Solid Primer</a></li>
```
justinwb, line 14: Should be _Solid Application Interoperability Primer_?

elf-pavlik: done

justinwb (approved):

.gitignore
```diff
@@ -1 +1 @@
-proposals/**/index.html
+proposals/*/index.html
```
ericprud, line 1: was that "**" intentional? was it intended to reach e.g. `proposals/foo/bar/index.html` ?

PR #70, opened 2020-11-10
Application access

elf-pavlik:
Initial PR to discuss app access


elf-pavlik:
@justinwb could you please take a look at Access Grant from the last commit https://github.com/solid/data-interoperability-panel/pull/70/commits/6dc3e85b9c51c55f6876099a819f039301df2170 ?

I would like to have it ready for Monday as an example of Alice giving read access to any Project and Task that she can access to Performchart. In that case we would look at possibility that Performchart has such grant as VC and presents it to `omni.example` RS when making request to projects and task it can discover using Alice's remote data registry.

elf-pavlik:
I wonder if we can merge this one and I would start updating primer to include changes from #71 as well as addressing #75 

Any problems this PR introduces we can track as issues

In a future I will be careful to separate renaming into dedicated PR

elf-pavlik:
I think we agreed to merge this one during the last call.

PR #81, opened 2021-01-28
application discovers public resources known by the user

elf-pavlik:
This UC tries to capture case where user doesn't get access receipt since resource is editable by public (or authorized agent). At the same time user wants to keep track of resources they have edited and expect new applications which work with those resources to discover them.

### DIFF
--- a/proposals/ucr/index.bs
+++ b/proposals/ucr/index.bs
@@ -201,10 +201,10 @@ grants Acme's customer group read acecss to analytic reports.
 ### Application can discover publicly accessible resources already known by the user
 
 Bob cultivates various interests and contributes to many independent wikis,
-including cocoascript.example, airwall.example, frogman.example etc.
-Bob have discovered new application to edit wiki pages wikitor.example,
+including cocoascript.example, airwall.example, frogman.example, etc.
+Bob discovers a new application to edit wiki pages, wikitor.example, and
 he authorizes it to read and write wiki pages on his behalf. Bob expects
-that wikitor will find all the wiki pages he alredy contributed to.
+that wikitor will find all the wiki pages to which he has already contributed.
 
 Requirements {#requirements}
 ================================================================================
--- a/proposals/ucr/index.bs
+++ b/proposals/ucr/index.bs
@@ -196,9 +196,9 @@ access to analytic reports, so it can distribute them to Acme customers.
 Acme also allows their customers to access analytic reports directly. Eric
 grants Acme's customer group read acecss to analytic reports.
 
-## Data Discovery
+## Data Discovery ## {#discovery}
 
-### Application can discover publicly accessible resources already known by the user
+### Application can discover publicly accessible resources already known by the user ### {#public-discovery}
 
 Bob cultivates various interests and contributes to many independent wikis,
 including cocoascript.example, airwall.example, frogman.example, etc.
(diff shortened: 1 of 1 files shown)


TallTed (changes requested):

TallTed:
Seems github had a hiccup, and only 1 of several suggestions went through yesterday. The others have gone through now.

proposals/ucr/index.bs
```diff
@@ -196,5 +196,15 @@ access to analytic reports, so it can distribute them to Acme customers.
 Acme also allows their customers to access analytic reports directly. Eric
 grants Acme's customer group read acecss to analytic reports.
 
+## Data Discovery
```
justinwb, line 199:
```suggestion
## Data Discovery ## {#discovery}
```

proposals/ucr/index.bs
```diff
@@ -196,5 +196,15 @@ access to analytic reports, so it can distribute them to Acme customers.
...
+
+### Application can discover publicly accessible resources already known by the user
```
justinwb, line 201:
```suggestion
### Application can discover publicly accessible resources already known by the user ### {#public-discovery}
```

justinwb (commented):
Is there some additional context here about where that data is stored? Specifically, is this public data that he controls, vs. data that he's created that is publicly available in a store that he doesn't? Possibly worth another use case?

joshdcollins:
@elf-pavlik Picking this up in the comments given the audio challenges today in our session.

Could this use case be achieved by defining a shapetree (let's call it Bookmark) for some data structure that tracks the links to the public data with any necessary contextual information?

For example, if I was using some editor application to update wikidata content, that application would request access to the data registration that aligns to the Bookmark shapetree.  Upon editing this public resource, the application updates the Bookmark resource with the new reference.

elf-pavlik:
I think we would need to consider how those bookmarks will differ from remote data registrations. We would also need to clarify when application uses remote data registrations and when it uses bookmarks. With bookmarks I think we would also need to evaluate if one application doesn't get knowledge about resources it should not know about, we shouldn't end up in situation where any application can access all the bookmarks.

joshdcollins:
In my mind, the intention of a remote data registration (RDR) is to detail all instances of a specific shapetree that have been explicitly shared with me (and would refer to data within another agent's data registration).

If the public data we are discovering exists within a public data registration I suppose it could referenced as a RDR, but would require a trusted agent to update that registry.  If the public data is not in a data registration (such as wikidata) I would lean towards an application storing those references as standard data within a data registration.

Authorizing access could applications could work as is currently defined with either all instances being authorized or specific instances.  If the metadata and contextual nature differs greatly between how applications use this data, it may tell us they are different shapetrees that should be authorized separately.

elf-pavlik:
> If the public data we are discovering exists within a public data registration I suppose it could referenced as a RDR, but would require a trusted agent to update that registry.

Possibly we would need to define how application could request that from trusted agent.


BTW with bookmarks how application would know which shape(tree) specific bookmarked resource conforms to? RDR already have that feature.

joshdcollins:
> BTW with bookmarks how application would know which shape(tree) specific bookmarked resource conforms to? RDR already have that feature.

I probably confused things using the term Bookmark.  In my mind that is just another shapetree (as opposed to a new primitive in the interop universe) -- some data structure that includes the link and some metadata.

PR #130, opened 2021-06-21
change shape path to referenced type and reference predicate

elf-pavlik:
I'm opening this draft PR just as a reference for what I should be able to implement Data Grant with `InheritsInstances` against. Actually I will test code using snippets from this branch.

Once we clarify how to expand prefixes in shape paths and possibly some other issues I will be able to update the implementation. I'll also create issues to track possible problems with using shape path.

We should also be able to discuss them tomorrow during panel meeting and on Wednesday during TS implementation meeting.

### DIFF
--- a/proposals/primer/index.bs
+++ b/proposals/primer/index.bs
@@ -61,7 +61,7 @@ All ShExC code snippets will asssume
   <pre class=include-code>
   path: snippets/solidshapes.example/shapes/project.shex
   highlight: shex
-  show: 7-100
+  show: 8-100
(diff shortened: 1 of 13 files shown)


elf-pavlik:
I think this one is ready to go, given that we alerady have sai-js working over those snippets. As soon as we resolve #131 I will update snippets in primer as well as sai-js implementation.

justinwb (approved):
Approving to continue momentum, though I think that we still need to settling the shape path discussion.

PR #137, opened 2021-07-02
~ use SPARQL version of prefixes in turtle

### LINKED ISSUE #125: use SPARQL version of prefixes in turtle
this will align it better with ShEx

ericprud:
closes #125,


elf-pavlik (approved):
awesome! i think we should just merge it @justinwb

Issue #132, opened 2021-06-28
Consider generalizing Application Registration to Agent Registration


elf-pavlik:
After initial work on implementing Remote Data Grants in [sai-js](https://github.com/janeirodigital/sai-js) I think that we should rethink how we manage Access Grants and Data Grants.

For Applications (Software Agents) we don't assume that they have persistent storage and we create Application Registration where we store their Access Receipt. On other hands for Social Agents we assume that they can receive Access Receipt in their inbox. This introduces requirement for maintaining copy of received Access Receipt and managing identifiers for it and its Data Grants. It becomes **especially** cumbersome for Data Grant with `interop:InheritInstances` scope which depends on another Data Grant and references it with `interop:inheritsFromGrant`. Furthermore we introduce another copy of those Data Grants for each relevant Remote Agent Data Registration, which references that copy with `interop:satisfiesDataGrant`. Having copy of a Data Grant attached to Access Receipt and later copy of a copy created for Remote Agent Data Registrations creates a lot of overhead in keeping track **especially** when we use `interop:inheritsFromGrant` relation between pair of Data Grants.

Moving to possible improvement which I haven't though of very thoroughly yet but I would like to invite us all to think of. Just as we create Application Registration which given Application (Software Agent) can access, we could generalize that pattern and create Agent Registration for Social Agents as well. Original Access Receipt and Data Grants (with stable `interop:inheritsFromGrant` relations) could be stored there. Social Agent receiving that Access Grant would have access to it and could reference relevant grants from Remote Agent Data Registration.

Furthermore, given that original Access Receipt gets stored in storage owned by the Social Agent who grants that access. We could reconsider just using Access Grant directly without creating a copy in form of Access Receipt. Last but not least, instead of denoting all data grants with fragment IRIs in the same document as Access Grant. We should consider describing each data grant in a separate document which should allow managing access to them separately.

elf-pavlik:
#149 should address this issue.

PR #141, opened 2021-07-19
Separate each Data Grant from Access Grant

elf-pavlik:
Each Data Grant now is described in separate resource. Access Grant and Access Receipt now references those grants instead of creating a copy. Also Remote Agent Data Registration simply references the original grant  - hosted by owner of data.
Avoiding copying grants makes things easier especially that `inheritsFromGrants` statements has to be included in both grants (in subject and object position), so we have two way linking.

I also introduced distinct `InheritRemoteInstances` scope for Remote Data Grant. It goes together with diagrams in
https://hackmd.io/s_w3eSF3RBSWY-OMhxPwig?view

With this version of snippets I should be able to wrap up first version of sai-js module for solid applications (non-trusted agents) and release it by the end of this month.


elf-pavlik:
related to #132

justinwb (approved):

PR #165, opened 2021-08-30
Use a singular registry set

### LINKED ISSUE #156: Replace registry sets with single resource
Should be a single document resource linked off of the Profile Document. Will include pointers to the following:

* Agent Registry
* Consent Registry
* Data Registries

justinwb:
Resolves #156. 

_Note: Also includes a bit of non-substantive cleanup of the table of contents._


elf-pavlik (approved):
I understand that we will still need to do another iteration on some of the sections changed here when we work on:

* Remote Data Registrations
* Access Receipts
* Trusted Grants

justinwb:
> I understand that we will still need to do another iteration on some of the sections changed here when we work on:
> 
> * Remote Data Registrations
> * Access Receipts
> * Trusted Grants

Yup that's the plan. Since we haven't made those changes yet I didn't want this submission to be in conflict with other parts of the document. Once those other changes are made we will end up with the example in #156

PR #176, opened 2021-09-09
Optimize referenced access needs and data grants

### LINKED ISSUE #124: Optimize referenced data grants and referenced access needs
I'm currently implementing Data Grants. I'm noticing possibility to make them a little bit more ergonomic. Looking at current draft we have:

image

I would find it more ergonomic to have

image

or even

image

We remove the indirection between Access Receipt and Referenced Data Grant, which is also a data grant itself. We still preserve relation between Referenced Data Grant and Data Grant it's inheritance depends on.

Thinking further, and keeping #118 in mind, I'm not sure if we really need Referenced Data Grant or we could simply use Data Grant with `InheritedInstances` scope. As discussed in #109 we would define specific shape for this scope (as well as all other scopes). In that case we would get nice and clear:

image

### LINKED ISSUE #140: Data Grant should include information about who owns the data
This is related to #132

Currently copies of Data Grants created for Remote Agent Data Registration (referenced with `interop:satisfiesDataGrant`) do not include information about who owns the data. If Remote Data Grant has scope `AllRemoteFromAgent` all `AllRemote`, one can keep track on who owns the data based on information from Remote Agent Data Registration. In case of `SelectedRemote` one doesn't have access at that level and can't tell who owns the data.

I think Data Grant should include information about who owns the data. Furthermore it might be beneficial that Remote Data Grant with scope `SelectedRemote` only can reference data grants from the same data owner. Also possibly including information directly on Remote Data Grant.

justinwb:
Removes explicit "Reference" classes per #124. Also was an opportune time to add the dataOwner to Data Grants per #140.

Resolves #124.
Resolves #140.

### DIFF
--- a/proposals/specification/access-grants.bs
+++ b/proposals/specification/access-grants.bs
@@ -223,10 +223,11 @@ valid for a [=Data Grant=]:
         each [=Data Instance=].
 * <b>`interop:InheritInstances`</b>
   *  When the [=Data Grant=] linked via
-     `interop:inheritsFromGrant` has a scope of `interop:SelectedInstances`, and one
-     or more [=Data Instances=] are selected, any inheriting [=Data Grants=]
+     `interop:inheritsFromGrant` has a scope of `interop:AllInstances`, or a scope
+     of `interop:SelectedInstances` with one
+     or more [=Data Instances=] selected, any inheriting [=Data Grants=]
      have the opportunity to narrow their own access scope to only include
-     "child" [=Data Instances=] linked to the selected "parent" [=Data Instances=].
+     "child" [=Data Instances=] linked to the "parent" [=Data Instances=].
   * The link must follow the [=Shape Tree Reference=] between
     the parent [=Data Grant=] shape tree and the inheriting [=Data Grant=]
     shape tree.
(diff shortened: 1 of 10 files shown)


proposals/specification/access-grants.bs
```diff
@@ -221,6 +221,18 @@ valid for a [=Data Grant=]:
         to the [=Data Grant=] via `interop:hasDataInstance`.
     * Permissions will be set for the [=Access Grant Subject=] on
         each [=Data Instance=].
+* <b>`interop:InheritInstances`</b>
+  *  When the [=Data Grant=] linked via
+     `interop:inheritsFromGrant` has a scope of `interop:SelectedInstances`, and one
```
elf-pavlik, line 226: why do we mention `interop:SelectedInstances` here? The parent Data Grant, linked via `interop:inheritsFromGrant`, could also have 'interop:AllInstances` scope.

elf-pavlik (approved):
+212 −486 👍

proposals/specification/access-grants.bs
justinwb: Good catch - have adjusted in most recent commit 👍

PR #193, opened 2021-09-30
Update access modes to be more specific

### LINKED ISSUE #184: Update access modes to be more granular
Move to acl:Read, acl:Update, acl:Create, acl:Delete, acl:Append

justinwb:
Resolves #184


proposals/specification/access-consents.bs
```diff
@@ -766,6 +845,59 @@ Jose's [=Agent Registry=].
...
+      <tr>
+        <td>`acl:Create`</td>
+        <td>Create new resources within the [=Data Instances=] where applicable. 
+        [=Access modes=] assigned to 
+        the creator on a created resource should be a union of 
+        `interop:accessMode` and `interop:createAccessMode` modes. 
+        <b>acl:Create is not currently supported. 
+        Substitute with acl:Write or acl:Append, but may exceed intended scope
+        of access.</b></td>
+      </tr>
```
elf-pavlik, line 875:
```suggestion
```

Currently, we don't support *Create* with *SelectedFromRegistry* scope. We probably should create issue to track or just add note that *Create* is not supported.

elf-pavlik (approved):
just one inline suggestion

proposals/specification/access-consents.bs
justinwb: @elf-pavlik In this context it would giving the grantee the ability to create resources within the Data Instance, not create new data instances on the data registration.

elf-pavlik: ok

PR #198, opened 2021-10-02
+ PREFIX decls for shex, acl and skos

ericprud:
tested in https://shexspec.github.io/ace-shexc-user/


elf-pavlik:
@ericprud can we remove interop-follow-refs.shex  and create it in github workflow to only be published on gh-pages?

ericprud:
Yep. On phone now (at a 2hr? show). Anyone who wants to jump on that move had my blessings

ericprud:
Also, I know nothing about GitHub workflows

justinwb (approved):

PR #204, opened 2021-10-06
Factor out API and Operations. Expand Data Authorization.

### LINKED ISSUE #179: Refactor API and Operations
It will be most efficient to do one full pass on the API and Operations after refactoring of authorization (access consents, grants) is completed.

justinwb:
Resolves #179.


elf-pavlik (approved):
I'm going to follow up with one or two issues after we merge it.

PR #219, opened 2021-11-23
Create 2021-11-23.md

elf-pavlik:


justinwb (approved):

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+- Server-side service used as an application by the user (e.g. user logs in and the user's webid is leveraged, and app needs to keep some state to know which to send for which user). Would be a different instance of fetch per user
+- Less explored - if the user is giving access to the organization the runs the application and the application is running "headless", when requests are made to the storage, the "organization" would appear as the social agent with the client identifier. Need to clarify who is being authorization - organization vs. user, and whose webid is presented.
+
+* Justin: I agree with how you laid out the two scenarios. It would be probably fine and appropriate to maintain the pairing as organization as social agent and application, when it operates in this 'headless' way. In the past the we were using same uri for webid and client_id. In application profile we already identify the responsible social agent as author.
+* ...: We have to detail out a little bit what the moving parts look on the authorization part on Solid-OIDC. Actually that's the first use case I need to demonstrate with Java library. I can take action item to dig into this one.
```
TallTed, line 65:
```suggestion
* Justin: I agree with how you laid out the two scenarios. It would be probably fine and appropriate to maintain the pairing as organization as social agent and application, when it operates in this 'headless' way. In the past, the we were using same URI for WebID and `client_id`. In application profile, we already identify the responsible social agent as author.
  * We have to detail out a little bit what the moving parts look on the authorization part on Solid-OIDC. Actually that's the first use case I need to demonstrate with Java library. I can take action item to dig into this one.
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+
+* Pavlik: if you could focus on the consent screen presented to the user, do they have an application registration, or is it a social agent registration? Possibly there will be cases for both. Possibly approach it in both ways.
+* Justin: My gut says that you would want to have Application Registration, but they could be case where we want to have Social Agent registration for the organization. I may want to authorize social agent services one by one.
```
TallTed, line 68:
```suggestion
* Justin: My gut says that you would want to have Application Registration, but there could be cases where we want to have Social Agent registration for the organization. I may want to authorize social agent services one by one.
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+
+* Eric: If there is pairing between 
+
+* Pavlik: When saying the user wants to authorize certain services, we could use access need groups for that. Services could be expressed that way. For different services there could be different access need groups.
```
TallTed, line 72:
```suggestion
* Pavlik: When saying the user wants to authorize certain services, we could use access need groups for that. Services could be expressed that way. For different services, there could be different access need groups.
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+
+* Justin: I think there is good reason to have both available. Organization could use access need groups under application umbrella. It's nice to have that flexibility
+
+### 
+
+* Eric: We can extend grant in the future that if grant has delegation privilege they can pass it further. As proof-carrying delegation. It's dynamic access control, owner doesn't maintain list of who has access.
```
TallTed, line 78:
```suggestion
* Eric: We can extend grant in the future such that if grant has delegation privilege, they can pass it further. As proof-carrying delegation. It's dynamic access control; owner doesn't maintain a list of who has access.
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+
+* Pavlik: Lines up with where i"m hoping to head with delegated / trusted grants. Already have setup for delegated grants between owner and application they use. Can use the same pattern in a chain. Also need to have an upstream notification of delegation. Omni gives Alice access. Alice delegates that access to Bob. Alice needs to let Omni know so it can update its records. Will share a link to authz panel.
```
TallTed, line 80:
```suggestion
* Pavlik: Lines up with where i"m hoping to head with delegated / trusted grants. Already have setup for delegated grants between owner and application they use. Can use the same pattern in a chain. Also need to have an upstream notification of delegation. Omni gives Alice access. Alice delegates that access to Bob. Alice needs to let Omni know, so it can update its records. Will share a link to authz panel. 
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+
+### 
+
+* Eric: We can extend grant in the future that if grant has delegation privilege they can pass it further. As proof-carrying delegation. It's dynamic access control, owner doesn't maintain list of who has access.
+
+* Pavlik: Lines up with where i"m hoping to head with delegated / trusted grants. Already have setup for delegated grants between owner and application they use. Can use the same pattern in a chain. Also need to have an upstream notification of delegation. Omni gives Alice access. Alice delegates that access to Bob. Alice needs to let Omni know so it can update its records. Will share a link to authz panel. 
+
+* Discussion: https://github.com/solid/authorization-panel/discussions/271
+
+* Eric: One strategy is proxying - where Alice never needs to know that Acme delegated. Could say that anyone can submit to me a proof chain and I'll just check it, and no one needs to be notified of those delegations.
```
TallTed, line 84:
```suggestion
* Eric: One strategy is proxying — where Alice never needs to know that Acme delegated. Could say that anyone can submit to me a proof chain, and I'll just check it; no one needs to be notified of those delegations.
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+* ...: We either need to say that Prescrito gets partial access to medical record, which would require something like shapemasks or tripple based authorization. Or we need to maintain those links in stand alone resources. This term of reference lists comes from that. We could say that Prescripto has access to prescriptions reference list but not to reference lists of conditions and appointments. It seems like the most straight forward approach. We would need to ensure that shape paths can follow external resources.
```
TallTed, line 98:
```suggestion
* Justin: https://hackmd.io/@hAlm_W2-T-q16baquNyRXg/HyS6Xuqdt
  * It gets the concept across. I used health scenario, since we deal with it a lot.
  * You have Data Registry and Data Registration for medical records. Medical record is composed of prescriptions and appointments and conditions. Medical record is more of a meta-object which just links to all those other things.
  * If I want to share the whole medical record with someone who helps taking care of me, I can share the whole medical record with them. To give partial access, such as when Alice wants to use Prescripto application, she wants it to access prescriptions, but not conditions, appointments, etc. Prescripto needs to see all the prescriptions, and also if it adds new one, it needs to link it to medical record. For Prescripto to know, it needs to be able to see those links. If Medical Record contains links to everything (conditions, appointments, prescriptions). Even if Prescripto can't follow those links, it still leaks data by knowing how many conditions and appointments are linked, which in itself is not acceptable.
  * We either need to say that Prescripto gets partial access to medical record, which would require something like shapemasks or triple-based authorization, or we need to maintain those links in standalone resources. The "reference lists" term comes from that. We could say that Prescripto has access to prescriptions reference list, but not to reference lists of conditions and appointments. It seems like the most straight forward approach. We would need to ensure that shape paths can follow external resources.
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+
+* Pavlik: Ties into what's the parent shape tree and the target shape tree (e.g. child). Needs to be a way in the medical record to say what the reference lists are for the children. With that information we'd be able to discover what the right links are.
+* Justin: I think the first place to investigate would be in shape tree definition. Shape trees are designed to provide that navigation of multi resource hierarchy. We already have the ground work for that. I can see where this notion of reference lists typo of a model could fit in.
```
TallTed, line 101:
```suggestion
* Pavlik: Ties into what's the parent shape tree and the target shape tree (e.g., child). Needs to be a way in the medical record to say what the reference lists are for the children. With that information we'd be able to discover what the right links are.

* Justin: I think the first place to investigate would be shape tree definition. Shape trees are designed to provide navigation of multi-resource hierarchy. We already have the ground work for that. I can see where this notion of a reference lists type of model could fit in.
```

meetings/2021-11-23.md
```diff
@@ -0,0 +1,115 @@
...
+* Eric: I think we could do it with shape masks.
+* Justin: I think shape masks are elegant long term.
+* Eric: When you want to PUT you need to know which part of the graph.
+* Pavlik: Might be useful to talk to GraphMetrix about triple level access control approach. Would be good to get some feedback from their implementation.
+* Justin: I believe that shape masks are so complementary to the rest of the stuff you need to do as application. It may be because we use shape trees in interop.
+* ...: I plan on expanding this use case a little bit more. I could try to design a shape tree that tries to map this.
+* Pavlik: What's important is to keep the reference list as a detail of the reference itself and not make some separate reference to it.
+* Justin: So - Encode the external reference (to the list) into the existing shape tree reference between medical record and prescription (Pavlik: ACK)
```
TallTed, line 109:
```suggestion
* Eric: When you want to `PUT`, you need to know which part of the graph.
* Pavlik: Might be useful to talk to GraphMetrix about triple-level access control approach. Would be good to get some feedback from their implementation.
* Justin: I believe that shape masks are complementary to the rest of the stuff you need to do as application. It may be because we use shape trees in interop.
  * I plan on expanding this use case a little bit more. I could try to design a shape tree that tries to map this.
* Pavlik: What's important is to keep the reference list as a detail of the reference itself and not make some separate reference to it.
* Justin: So — Encode the external reference (to the list) into the existing shape tree reference between medical record and prescription? (Pavlik: ACK)
```

TallTed (changes requested):
Minor tweaks to markup, punctuation, typos.

Issue #210, opened 2021-10-19
Incorporate profile requirements from solid-oidc for Social Agents and Applications.


justinwb:
The specification should do a better job incorporating requirements of Social Agents and Applications as stipulated by [Solid-OIDC](https://solid.github.io/solid-oidc/):

- Incorporate oidcIssuer into the Social Agent profile examples
- Reconcile [Client Identifier Document](https://solid.github.io/solid-oidc/#clientids-document) and [Application](https://solid.github.io/data-interoperability-panel/specification/#app) Profile

PR #231, opened 2022-01-27
Add jsonld context for application profile

justinwb:
Adding a jsonld context specifically for the [Application Profile](https://solid.github.io/data-interoperability-panel/specification/#app), which needs to be able to cross-pollinate with the contents of the [Solid-OIDC Client Identifier Document](https://solid.github.io/solid-oidc/#clientids-document) (see https://github.com/solid/data-interoperability-panel/issues/210).

There are a couple of determinations we still need to make which would adjust the contents of the `@context` in this PR:

- Rather than including `application_name`/`interop:applicationName` in this context, should we instead use the value from `oidc:client_name`? 
- Rather than including `application_thumbnail`/`interop:applicationThumbnail` in this context, should we instead use the `oidc:logo_uri`?
- Would `oidc:contacts` be equivalent to `application_author`/`interop:applicationAuthor`?

Personally I lean towards DRY and feel we should reuse what's already being specified in solid-oidc and extend


elf-pavlik:
Looking at https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata

> `logo_uri` - OPTIONAL. URL that references a logo for the Client application. If present, the server SHOULD display this image to the End-User during approval. The value of this field MUST point to a valid image file.

> `client_name` - OPTIONAL. Name of the Client to be presented to the End-User.

> `contacts` - OPTIONAL. Array of e-mail addresses of people responsible for this Client. This might be used by some providers to enable a Web user interface to modify the Client information. 

Since `contacts` is defined as array of e-mail addresses we probably don't want to use it. The other two seem appropriate to me.

proposals/specification/interop-application.jsonld
```diff
@@ -0,0 +1,28 @@
...
+      "@id": "http://www.w3.org/ns/solid/interop#applicationDescription"
+    },
+    "application_author": {
+      "@id": "http://www.w3.org/ns/solid/interop#applicationAuthor",
+      "@type": "@id"
+    },
+    "application_thumbnail": {
+      "@id": "http://www.w3.org/ns/solid/interop#applicationThumbnail",
+      "@type": "@id"
+    },
```
elf-pavlik, line 18:
```suggestion
```

proposals/specification/interop-application.jsonld
```diff
@@ -0,0 +1,28 @@
+{
+  "@context": {
+    "@version": 1.1,
+    "@protected": true,
+    "application_name": {
+      "@id": "http://www.w3.org/ns/solid/interop#applicationName"
+    },
```
elf-pavlik, line 7:
```suggestion
```

elf-pavlik (approved):

justinwb:
> The other two seem appropriate to me.

👍  Agree - so now the question becomes whether we should remove those terms from the interop vocabulary, adjust the current spec to reflect the terms we are using, etc. Thoughts?

PR #233, opened 2022-02-03
Add contexts and agent profiles for json-ld

justinwb:
Created a sub-folder to house json-ld contexts since there will be more. Renamed to application-profile context and added another for social agent profile.


proposals/specification/contexts/social-agent-profile.jsonld
```diff
@@ -0,0 +1,26 @@
...
+    },
+    "access_inbox": {
+      "@id": "http://www.w3.org/ns/solid/interop#hasAccessInbox",
+      "@type": "@id"
+    },
+    "registry_set": {
+      "@id": "http://www.w3.org/ns/solid/interop#hasRegistrySet",
+      "@type": "@id"
+    },
+    "oidc_issuer": {
```
acoburn, line 17: Why `oidc_issuer` and not `oidcIssuer`? That seems atypical. Same applies to the other properties.

elf-pavlik:
@justinwb could you please paste in your initial comment an example of document using this context together with Solid-OIDC one?

proposals/specification/contexts/social-agent-profile.jsonld
justinwb: no strong preference one way or the other. most json-ld that I've seen uses this style, including solid-oidc context, so I stuck with that. not against camelcasing.

justinwb:
> @justinwb could you please paste in your initial comment an example of document using this context together with Solid-OIDC one?

@elf-pavlik I think you mean the thread from this PR? https://github.com/solid/data-interoperability-panel/pull/231

elf-pavlik:
I mean JSON-LD snippet which would replace https://solid.github.io/data-interoperability-panel/specification/snippets/projectron.example/projectron.example.ttl

I understand that this context is intended to layer on top of solid-oidc one for Client ID documents.

elf-pavlik:
I think examples of using both contexts would be helpful. I assume that you went with snake case to match solid-oidc context.

justinwb:
> I think examples of using both contexts would be helpful. I assume that you went with snake case to match solid-oidc context.

Switched to camelCase for social agent profile. Not sure if we should do the same for application profile given the cross-pollination with the client identifier doc from solid-oidc which doesn't use it.

acoburn:
I'm much happier with the camelCase format, thanks! 

> most json-ld that I've seen uses this style, including solid-oidc context

To be clear, solid-oidc uses the underscore style for the simple reason that it aims to align with https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata

justinwb:
> To be clear, solid-oidc uses the underscore style for the simple reason that it aims to align with https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata

@acoburn feels like it may be worth leaving the interop application context with the same style for the same reason since they'd be interspersed. will use camelCase on the rest 🙂 

While you're here - it'd be great align on a best practice for a client id document incorporating both contexts. Here's what I've been doing for the time being: 


```
  {
    "@context": [
        "https://www.w3.org/ns/solid/oidc-context.jsonld",
        "https://solid.github.io/data-interoperability-panel/specification/interop-application.jsonld"
    ],
    "client_id": "https://projectron.example/id",
    "client_name": "Projectron",
    "redirect_uris": [ "https://projectron.example/redirect" ],
    "post_logout_redirect_uris": [ "https://projectron.example/logout" ],
    "client_uri": "https://projectron.example/",
    "logo_uri": "https://projectron.example/logo.png",
    "tos_uri": "https://projectron.example/tos.html",
    "scope": "openid offline_access profile",
    "grant_types": [ "refresh_token", "authorization_code" ],
    "response_types": [ "code" ],
    "default_max_age": 3600,
    "require_auth_time": true,
    "application_description": "Best project management ever",
    "application_author": "https://acme.example/id",
    "has_access_need_group": [ "https://projectron.example/access#group1", "https://projectron.example/access#group2" ]
  }
```

[Example live on JSON-LD playground](https://tinyurl.com/26ba8nrc)

acoburn:
Looks good!

Only (minor) comment is to include `webid` in the scope field

justinwb:
> Looks good!
> 
> Only (minor) comment is to include `webid` in the scope field

Ah great catch. I actually had to remove that from my local instance because I was getting rejections from CSS' IDP if I provided it (although it proceeds to issue tokens with it). Good reminder to file an issue for it!

elf-pavlik (approved):

PR #239, opened 2022-02-09
Fix AccessNeedDescription pointing back to AccessNeedGroup

angelaraya:
The shape and description table was pointing to AccessNeedGroup instead of AccessNeed. The snippet appears to be correct.


elf-pavlik (approved):
Thank you @angelaraya 
I just fixed one line in the snippet since you didn't include it in this PR.

justinwb (approved):

PR #245, opened 2022-02-17
Update JSON-LD context with access consents, data consents, access grants, data grants

justinwb:
Adds support for [Access Consent Registry](https://deploy-preview-244--data-interoperability-panel.netlify.app/specification/#access-consent-registry), [Access Consents](https://deploy-preview-244--data-interoperability-panel.netlify.app/specification/#access-consent), [Data Consents](https://deploy-preview-244--data-interoperability-panel.netlify.app/specification/#data-consent), [Access Grants](https://deploy-preview-244--data-interoperability-panel.netlify.app/specification/#access-grant), [Data Grants](https://deploy-preview-244--data-interoperability-panel.netlify.app/specification/#data-grant), [Delegated Data Grants](https://deploy-preview-244--data-interoperability-panel.netlify.app/specification/#delegated-data-grant) to interop JSON-LD context.

Open question / note - for consistency I added another entry for "hasAccessNeedGroup", though we have a snake case version (has_access_need_group) that is specifically used in the client identifier document. We use this term in many other places, and seems like it would be nice to have it be consistent with others (camelCased) in those places. That said, what is the best practice in JSON-LD for a case like that?

### DIFF
--- a/proposals/specification/interop.jsonld
+++ b/proposals/specification/interop.jsonld
@@ -122,7 +122,7 @@
       "@id": "interop:scopeOfGrant",
       "@type": "@id"
     },
-    "satisfiesAccessNed": {
+    "satisfiesAccessNeed": {
       "@id": "interop:satisfiesAccessNed",
       "@type": "@id"
     },


proposals/specification/interop.jsonld
```diff
@@ -86,6 +74,133 @@
...
+      "@container": [
+        "@id",
+        "@set"
+      ]
+    },
+    "scopeOfGrant": {
+      "@id": "interop:scopeOfGrant",
+      "@type": "@id"
+    },
+    "satisfiesAccessNed": {
```
elf-pavlik, line 133:
```suggestion
    "satisfiesAccessNeed": {
```

proposals/specification/interop.jsonld
```diff
@@ -86,6 +74,133 @@
...
+      "@type": "@id"
+    },
+    "hasAccessNeedGroup": {
+      "@id": "interop:hasAccessNeedGroup",
+      "@type": "@id",
+      "@container": [
+        "@id",
+        "@set"
+      ]
+    },
```
elf-pavlik, line 96: I think we should only keep one otherwise I'm not sure which one gets picked up during compacting.
To be honest I think we should just dish the camel case ones and accept mixed casing patterns in a single document.

elf-pavlik: Actually, I meant to say that we should dish the snake case version 🐍. Should have imagine the animal before posting 🐫 .

Gland no kebab cases involved so far 🍢 😅

Issue #243, opened 2022-02-16
Access Needs for Public Resources


laurensdeb:
## Suggested Use Case
Suppose I build an application allowing a user to manage a website and serve it from their Solid Pod. Supposing we have some way of storing such a static website in a permissive tree (see #240), and are able to describe an access need for this data. 

## Open questions
- How would an application then define an access need for making this instance of the permissive tree public?
- Also, it might be the case that I want to serve multiple websites, from different instances of such a permissive tree. Would I send an [Access Request](https://solid.github.io/data-interoperability-panel/specification/#access-request) for this purpose?

## Suggested course of action

- Define the concept of a public agent, like `foaf:Agent` is to be used in WAC for this purpose.
- Perhaps the latter question should actually be clarified under the "Social Agent Shares Access"-flow that is already mentioned in the specification.

PR #254, opened 2022-03-15
Access Needs for Public Resources

laurensdeb:
## Context
Some applications may have to express the need for making some data instances available to public agents, thus the interoperability specification should allow for applications to define access modes for public agents in their access needs. We detailed this requirement in #243.

## In progress
- [x] Add `interop:publicAccessMode` to Access Needs,  Data Grant and Data Consent in interop.ttl, JSON-LD contexts and specification.
- [ ] Further explain the behaviour of this field in the specification
- [ ] Update some of the snippets to include this predicate (to be discussed in next panel)


elf-pavlik:
I think we need to clarify
1. What value would we use for `interop:grantee`?
2. Where do we store generated AccessGrant?

Based on the above I will have a better understanding of the approach to see how the new property would be used.

@laurensdeb could you please prioritize snippets with examples? I think this will help a lot with discussing the proposed approach.

laurensdeb:
1. The idea behind the proposed `interop:publicAccessMode` is that it expresses the resources the agent wishes to make public, i.e. the `interop:grantee` of such a DataGrant would still be the agent that has expressed this Access Need.

2. The AccessGrant would thus be stored under the AgentRegistry of the agent that expressed the access need.

An example of a Data Grant:

```
alice-projectron:40d038ea
  a interop:DataGrant ;
  interop:dataOwner alice:\#id ;
  interop:grantee projectron:\#id ;
  interop:registeredShapeTree pm-shapetrees:ProjectTree ;
  interop:hasDataRegistration alice-work-data:8501f084\/ ;
  interop:satisfiesAccessNeed projectron:\#ac54ff1e ;
  interop:accessMode acl:Read, acl:Create ;
  interop:creatorAccessMode acl:Update, acl:Delete ;
  interop:publicAccessMode acl:Read;
  interop:scopeOfGrant interop:AllFromRegistry .
  ```
  
** One thing I'm not sure about is how this would work in the "Social Agent Shares Access"-flow, supposing that an end-user wishes to share a resource to the public outside of interaction with an application or a request from a concrete social agent. This might be a use case we want to support.**

ericprud:
> **One thing I'm not sure about is how this would work in the "Social Agent Shares Access"-flow, supposing that an end-user wishes to share a resource to the public outside of interaction with an application or a request from a concrete social agent. This might be a use case we want to support.**

To make sure I understand, would a FOAF page or resumé be an example of a public page not associated with an app or social agent?

elf-pavlik:
We should try to take it into account together with #291
Some of my feedback there relates supporting discovery of data shared with the public.

PR #290, opened 2022-12-05
Meeting minutes 2022-12-05

elf-pavlik:


meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
+# 2022-12-05 Solid Interoperability
```
TallTed, line 1: from [README.md](https://github.com/solid/data-interoperability-panel/blob/main/README.md#meetings)
> Panel meetings are held most Tuesdays at 10:00AM Eastern / 4:00PM Central European. Check-in on [gitter](https://gitter.im/solid/data-interoperability-panel) for weekly agenda.

It's too bad today isn't a Tuesday, and no "agenda" was committed until it was these "minutes".

Today's [Gitter log](https://gitter.im/solid/data-interoperability-panel) suggests the [README.md](https://github.com/solid/data-interoperability-panel/blob/main/README.md#meetings) should be updated to reflect "every other Monday" instead of "most Tuesdays".

Also, the [Solid CG calendar](https://www.w3.org/groups/cg/solid/calendar) DESPERATELY needs updates to include all regularly scheduled meetings (for whole CG, specific panels, just editors, etc.). Further, it _should_ be possible to add special meetings as soon as they're scheduled.

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Minutes
+
+
+### Agent-Specific Discovery
+
+Issue: https://github.com/solid/data-interoperability-panel/issues/289
+Spec: https://woutermont.github.io/solid/agent-specific-discovery.html
+
+* WT: I worte it based on arguments I put forwrad. To better align with different approaches to interop it might be better to split different aspects in separate specs. This is my first attempt, taking most of the agent registration aspects.
```
TallTed, line 36:
```suggestion
* WT: I wrote it based on arguments I put forward. To better align with different approaches to interop, it might be better to split different aspects into separate specs. This is my first attempt, taking most of the agent registration aspects.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* ...: Introduction tries to explain why this spec is needed. Showing that try and error approach is not feasible. The current draft of interop spec takes the identity of client and points to agent registration. This step is valuable and important piece of the puzzle so we can define it separatelly.
```
TallTed, line 37:
```suggestion
* ...: Introduction tries to explain why this spec is needed, showing that trial-and-error approach is not feasible. The current draft of the interop spec takes the identity of the client and points to agent registration. This step is a valuable and important piece of the puzzle so we can define it separately.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* ...: It should be possible to incorporate resource discovery hub in other services. I gave an example of OIDC Provider but should also work with authorization agent.
```
TallTed, line 38:
```suggestion
* ...: It should be possible to incorporate a resource discovery hub into other services. I gave an example of an OIDC Provider but it should also work with an authorization agent.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* ...: The idea is that client who wants to discover resource discovery hub, dereferences the id to get the document. We define link headers and we can fallback to statements in the document. We also define uniqueness of discovery hub to the entity. If you have the discovery hub iri the client can call it for specific information. It happens in the same way that agent registration discovery happens in current SAI spec.
```
TallTed, line 39:
```suggestion
* ...: The idea is that a client who wants to discover a resource discovery hub, dereferences the `id` to get the document. We define link headers and we can fall back to statements in the document. We also define uniqueness of discovery hub to the entity. If you have the discovery hub IRI, the client can call it for specific information. It happens in the same way that agent registration discovery happens in the current SAI spec.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* ...: The last part is about agent registration, where agent can request a agent specific document at the discovery hub.
```
TallTed, line 40:
```suggestion
* ...: The last part is about agent registration, where an agent can request an agent-specific document at the discovery hub.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* ...: There is a sequence diagram which follows all the steps.
+
+* Pavlik: Question about registration. The way I implemented it - the owner of Authz Agent handles it - the client redirects to the Authz Agent which handles it. So some user interaction is required.
+* WT: Need a way to register agents but don't stipulate the methods to do so. This spec just indicates the possibility. The basic registration proposed in the doc is to illustrate how back and forth could work but leaves implementation open to which type of registration is stipulated.
+* Pavlik: How in practice would you see it with a PR? Extract parts of SAI and then SAI extends more specifically?
+* WT: I tried to mention / use the SAI vocab in here - so parts of SAI that are similar could be left out, and then decide whether to change vocab here to reflect original vocab. Don't think it's necessary to have general and more specific - just make one that's standalone and in rest of SAI extend with specific implementation of registration procedures
```
TallTed, line 46:
```suggestion
* WT: I tried to mention / use the SAI vocab in here - so parts of SAI that are similar could be left out, and then decide whether to change vocab here to reflect original vocab. Don't think it's necessary to have general _and_ more specific — just make one that's standalone, and in rest of SAI extend with specific implementation of registration procedures
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* Pavlik: Currently in Sai the Authz Agent handles discovery of Agent Registration. It has logic based on client and end user identities to return the right registration. Would you see this part defined here or in SAI
+* WT: Could include here, or could be implementation direction
+* Pavlik: Discussing possibility of having a special public agent registration and IRI for public agent (I think ACP does this). If we have something like that - maybe there would be two link headers returned for public and agent specific - or add a special triple to shortcut the hub - because you won't need to request authorization for public agent
+* WT: If you're a public agent and you act as such - and are registered nowhere - then you could nevertheless pull from the RDH to get the public agent location, or could use a specific triple in the webid. But using this mechanism you can separate technical stuff from the webid and leave it open for read/write of profile or whatever else you want to do with it.
```
TallTed, line 50:
```suggestion
* WT: If you're a public agent and you act as such — and are registered nowhere — then you could nevertheless pull from the RDH to get the public agent location, or could use a specific triple in the WebID. But using this mechanism you can separate technical stuff from the WebID and leave it open for read/write of profile or whatever else you want to do with it.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* Pavlik: I wouldn't say the client is acting as a public agent - any agent is always acting as a public agent plus an authorized agent. So you can get agent specific one plus public one. If you wanted one entry point it would be useful to always get the public registration as a separate link in a header. Advertising in a webid document could be seen as a shortcut (nice-to-have). Could also have a different predicate to advertise it. Would still have a structure of agent registration. To create it you'd still want to use an authorization agent.
```
TallTed, line 51:
```suggestion
* Pavlik: I wouldn't say the client is acting as a public agent — any agent is always acting as a public agent plus an authorized agent. So you can get agent-specific one plus public one. If you wanted one entry point, it would be useful to always get the public registration as a separate link in a header. Advertising in a WebID document could be seen as a shortcut (nice-to-have). Could also have a different predicate to advertise it. Would still have a structure of agent registration. To create it, you'd still want to use an authorization agent.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* Pavlik: In general I think it's an interesting approach to extract something but need to see how to recombine it. Definitely worth exploring.
+* Justin: I agree. One of the criticisms is that is to big, and it has gotten in the way of adoption. Having it more modular and havig pieces that can be adopted one at the time. I think part of creating somthing like this must to be paired with what changes will need to be done in the spec document.
+* ...: The current spec is imlemented so we know it works, if we extract one and we update the main document to complement it we should know that it will stil work.
+* ...: I think we still should mention how to use type index or migrate from it. Showing the pathway or a bride is important.
+* WT: I think so and this is the main reason I started doing this. The quesiton is very importan about what is this Public Agent Registration. Do we envision it as something separate again? Or do we say that this could be a document pointed by a discovery hub, but is not unimaginable that it would be the webid itself.
```
TallTed, line 56:
```suggestion
* WT: I think so, and this is the main reason I started doing this. The very important question is, what is this Public Agent Registration? Do we envision it as something separate again? Or do we say that this could be a document pointed to by a discovery hub, but is not unimaginable that it would be the WebID itself?
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* Justin: I really appreciate the you wrote this up. We've gotten a lot of feedback and suggestions, but not always follow up with contributing material.
+* WT: I can prepare PRs with this draft and changes to current spec.
+* Pavlik: We should think how we want to break up the specs and how dependency graph will look.
+* Pavlik: For agent registrations we have Application and Social Agent, we also discuss Public Agent. 
+* WT: In near feature I think this will suffice
```
TallTed, line 61:
```suggestion
* WT: I think this will suffice for near feature.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* Pavlik: Can you already identify something?
+* Pavlik: I can only imagine some VC based authorization that doesn't rely on identiy of the end user or the client.
+* WT: Yes as I said we can have something in the future, for example zaps-ld
```
TallTed, line 64:
```suggestion
* WT: Yes, as I said, we can have something else in the future; for example, `zaps-ld`.
```

meetings/2022-12-05.md
```diff
@@ -0,0 +1,91 @@
...
+* 
+
+### Type indexes
+
+* eP: I'll prepare proposal for meeting in next two weeks with position on type indexes
```
TallTed, line 69:
```suggestion
* eP: I'll prepare proposal for next meeting in two weeks, with position on type indexes.
```

TallTed (changes requested):
The usual typo-fixes and such

PR #300, opened 2023-02-27
minor fixes of README.md

TallTed:


elf-pavlik (approved):

elf-pavlik:
The issue causing the build to fail was addressed as part of #297 we can ignore the failing check on netlify.

Issue #295, opened 2023-02-12
2023-02-13 Meeting Agenda


elf-pavlik:
Meeting details in the [README](https://github.com/solid/data-interoperability-panel#meetings)
Please RSVP with :+1: /  :-1: reaction
Conclusions will be captured directly in each discussed issue or PR.
Please propose additional PRs and Issues by commenting.

### Pull Requests

* #291
* #254 

### Issues

* #279 
* #294 
* #293 & #278
* #280 
* #224

TallTed:
This agenda should include the jitsi details, as well as any other links that will help people join & participate....

------

PR #291 is a monster. I'm about halfway through it, and am certain I'll have more suggestions for the latter half. I can submit my partial review now if desired, or I can continue reviewing after the concall.

elf-pavlik:
I added link to [Meetings section in the README](https://github.com/solid/data-interoperability-panel#meetings)

Since we are going to meet in 30min I wouldn't expect people to have a chance to read your review before meeting starts. I think you can do it either way, submit partial review now or just finish it and submit it after the meeting.

PR #301, opened 2023-02-27
[specification] drop inboxes

### LINKED ISSUE #280: Remove public inboxes
Related issue: https://github.com/solid/specification/issues/464

Currently, spec suggest two inboxes in https://solid.github.io/data-interoperability-panel/specification/#social-agents

```ttl
  ######## Inboxes ########
  interop:hasInbox alice-inbox:general ;
  interop:hasAccessInbox alice-inbox:access .
```

Later on, Access Inbox is being used in https://solid.github.io/data-interoperability-panel/specification/#access-receipt

I think we should remove the use of public inboxes altogether. Instead, iterate on [SAI Invitations](https://github.com/solid/data-interoperability-panel/tree/main/proposals/invitation). I have some thoughts captured in [Social Agent Invitation and Identity Verification](https://hackmd.io/gIujZEXNROunumZkxn_pAg). I believe that we can leverage a pre-established communication channel between Social Agents to establish initial pair of Social Agent Registrations. Removing the need for any public inboxes altogether.

This will auto fix #227 😉

elf-pavlik:
closes #280

as agreed during #295

### DIFF
--- a/proposals/specification/diagrams/social-agent-shares-access.seq.mmd
+++ b/proposals/specification/diagrams/social-agent-shares-access.seq.mmd
@@ -28,9 +28,7 @@ sequenceDiagram
     Note over aa_alice, pod_alice: 14. Record Access Authorization
     Note over aa_alice, pod_alice: 15. Create Social Agent Registration
     Note over aa_alice, pod_alice: 16. Record Access Grant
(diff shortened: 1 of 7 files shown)


justinwb (approved):
Happy to eliminate these 🙂

PR #303, opened 2023-02-27
immutable Access Need Group and Access Need

### LINKED ISSUE #279: Define Access Need and Access Need Group as Immutable
We already do it for Authorizations (Access & Data) as well as Grants (Access & Data).

Having it also for Access Need, and Access Need Group will minimize the chance of any discrepancy between Authorization / Grant and Need it satisfies.

It will also allow the Authorization to use super simple caching of all the needs that were used for generating authorizations and grants.

I don't see a reason for when Access Need and Access Need group changes, they can't get a new IRI.
We can look at *supersedes* links for cases that want to keep the provenance trace.

### TODO

* [ ] #177

elf-pavlik:
closes #279


justinwb (approved):
✅

PR #304, opened 2023-02-27
added grantedBy to Data Authorization

### LINKED ISSUE #246: Add grantedBy to DataAuthorization
Implementation experience is that this would be useful to have within the DataConsent rather than needing to source it from the AccessConsent.

elf-pavlik:
closes #246

### DIFF
--- a/proposals/specification/snippets/alice.example/8fac3576.ttl
+++ b/proposals/specification/snippets/alice.example/8fac3576.ttl
@@ -10,8 +10,8 @@ PREFIX alice-sarah: <https://alice.example/agents/b49afcdf/>
 
 alice-sarah:8fac3576
   a interop:DataGrant ;
-  interop:dataOwner alice:\# ;
-  interop:grantedBy alice:\# ;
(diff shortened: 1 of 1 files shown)


elf-pavlik:
We should also consider relying on *creator* information tracked by the server. In a way we duplicate it here.

proposals/specification/snippets/alice.example/8fac3576.ttl
```diff
@@ -10,7 +10,8 @@ PREFIX alice-sarah: <https://alice.example/agents/b49afcdf/>
 
 alice-sarah:8fac3576
   a interop:DataGrant ;
-  interop:dataOwner alice: ;
+  interop:dataOwner alice:\# ;
+  interop:grantedBy alice:\# ;
```
srosset81, line 14:
```suggestion
  interop:dataOwner alice:\#id ;
  interop:grantedBy alice:\#id ;
```

srosset81 (approved):

